PR #1, opened 2022-01-24
Supervisor setup with 2 dumb processes

grejdi-mbta:
### In Action

See it in action with these steps:

1. Build container:
```sh
# for Podman, substitute 'docker' for 'podman'
docker build --no-cache -t data_platform_main__local -f docker/Dockerfile.main .
```

2. Run container:
```sh
# for Podman, substitute 'docker' for 'podman'
docker run --rm --volume $PWD:/data_platform:ro data_platform_main__local supervisord -n -c /data_platform/supervisor/conf/dataplatform.conf
```

3. Stop container with CTRL-C. (Note: Container will not stop right away.)

### Links

* [Asana](https://app.asana.com/0/1201701089427502/1201701089427505/f)

### DIFF
--- a/data_platform/__init__.py
+++ b/data_platform/__init__.py
@@ -1,0 +1,18 @@
+
+import os
+import logging
+from dotenv import load_dotenv
+
+
+# get enviroment variables from .env file, if not already set
+load_dotenv()
+
+# configure logging
+logLevel = logging.ERROR
+if os.environ.get('ENV', '') == 'local':
+  logLevel = logging.INFO
+# config
+logging.basicConfig(
+  format='%(asctime)s %(levelname)s %(message)s',
+  level=logLevel
+)
--- a/data_platform/batch/cubic_ods/process_incoming.py
+++ b/data_platform/batch/cubic_ods/process_incoming.py
@@ -1,0 +1,13 @@
+
+import logging
+import time
+
+
+def run():
+  # sample process
+  logging.error('START process_incoming')
+  time.sleep(10)
+  logging.error('END process_incoming (Took 10 seconds)')
+
+if __name__ == '__main__':
+  run()
--- a/supervisor/conf/dataplatform.conf
+++ b/supervisor/conf/dataplatform.conf
@@ -1,0 +1,42 @@
+
+[unix_http_server]
+file=/tmp/supervisor.sock
+
+[supervisord]
+logfile=/tmp/supervisord.log
+logfile_maxbytes=50MB
+logfile_backups=10
+loglevel=info
+pidfile=/tmp/supervisord.pid
+nodaemon=false
+silent=false
+minfds=1024
+minprocs=200
+
+[rpcinterface:supervisor]
+supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
+[supervisorctl]
+serverurl=unix:///tmp/supervisor.sock
+
+; -------------------------------------------------------------------------------------------------
+
+[program:cubic_ods__process_incoming]
+command=/usr/local/bin/python -m supervisor.cubic_ods__process_incoming
+directory=/data_platform
+autostart=true
+autorestart=true
+stopwaitsecs=300 ; wait 5 minutes before killing outright
+redirect_stderr=true
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+
+[program:cubic_ods__start_ingestion]
+command=/usr/local/bin/python -m supervisor.cubic_ods__start_ingestion
+directory=/data_platform
+autostart=true
+autorestart=true
+stopwaitsecs=300 ; wait 5 minutes before killing outright
+redirect_stderr=true
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
--- a/supervisor/cubic_ods__process_incoming.py
+++ b/supervisor/cubic_ods__process_incoming.py
@@ -1,0 +1,26 @@
+
+import signal
+
+from data_platform.batch.cubic_ods import process_incoming
+
+
+def run():
+  # kill switch state
+  kill = {
+    'cubic_ods__process_incoming': False
+  }
+  # kill switch handler
+  def sigtermHandler(signum, frame):
+    kill['cubic_ods__process_incoming'] = True
+
+  # register handlers for termination signals, so we don't kill in mid process
+  signal.signal(signal.SIGINT, sigtermHandler) # SINGINT 2 (ctrl-c)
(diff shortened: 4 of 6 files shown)


data_platform/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# get enviroment variables from .env file, if not already set
+load_dotenv()
+
+# configure logging
+logLevel = logging.ERROR
+if os.environ.get('ENV', '') == 'local':
+  logLevel = logging.INFO
+# config
+logging.basicConfig(
+  format='%(asctime)s %(levelname)s %(message)s',
```
grejdi-mbta, line 16: This format conforms with supervisor's format as well.

data_platform/batch/cubic_ods/process_incoming.py
```diff
@@ -0,0 +1,13 @@
+
+import logging
+import time
+
+
+def run():
+  # sample process
+  logging.error('START process_incoming')
+  time.sleep(10)
```
grejdi-mbta, line 9: Simulating some time passing.

supervisor/conf/dataplatform.conf
```diff
@@ -0,0 +1,42 @@
+
```
grejdi-mbta, line 1: See PR Description for how to test running this container.

supervisor/cubic_ods__process_incoming.py
```diff
@@ -0,0 +1,26 @@
+
```
grejdi-mbta, line 1: This setup allows for gracefully stopping the process.

supervisor/conf/dataplatform.conf
```diff
@@ -0,0 +1,42 @@
...
+serverurl=unix:///tmp/supervisor.sock
+
+; -------------------------------------------------------------------------------------------------
+
+[program:cubic_ods__process_incoming]
+command=/usr/local/bin/python -m supervisor.cubic_ods__process_incoming
+directory=/data_platform
+autostart=true
+autorestart=true
+stopwaitsecs=300 ; wait 5 minutes before killing outright
```
grejdi-mbta, line 29: This is how long supervisor waits to gracefully stop. It then issues a kill command.

paulswartz (commented):
- question: is `supervisor/` a Python module, or a directory for scripts? If it's the latter, I don't believe you need a `__init__.py`.
- when I used `^C` to exit out of the container, it exited immediately: it didn't appear to wait for anything to shut down.

Overall, this seems like a lot of boilerplate for something that's pretty straightforward in Elixir, and it would match our other ECS applications. Given that this application isn't interfacing directly with Spark (or even the data at all), does it make sense for it to be in Python?

supervisor/conf/dataplatform.conf
```diff
@@ -0,0 +1,42 @@
+
+[unix_http_server]
+file=/tmp/supervisor.sock
+
+[supervisord]
+logfile=/tmp/supervisord.log
```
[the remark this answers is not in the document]
grejdi-mbta: No need. Supervisor also outputs to stdout. You should have seen this (also in the log file) when running the container:
```
2022-01-24 17:30:38,222 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.
2022-01-24 17:30:38,226 INFO RPC interface 'supervisor' initialized
2022-01-24 17:30:38,226 CRIT Server 'unix_http_server' running without any HTTP authentication checking
2022-01-24 17:30:38,227 INFO supervisord started with pid 1
2022-01-24 17:30:39,233 INFO spawned: 'cubic_ods__process_incoming' with pid 9
2022-01-24 17:30:39,239 INFO spawned: 'cubic_ods__start_ingestion' with pid 10
```

data_platform/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+from dotenv import load_dotenv
+
+
+# get enviroment variables from .env file, if not already set
+load_dotenv()
+
+# configure logging
+logLevel = logging.ERROR
+if os.environ.get('ENV', '') == 'local':
+  logLevel = logging.INFO
```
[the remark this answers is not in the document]
grejdi-mbta: For log levels, I've felt that there are only 2 type of environments, local and not local. :)
So in general I prefer leaving the log level out of the config. I'd like to leave it out for now, and come back to it when we have the use case for it.

supervisor/conf/dataplatform.conf
grejdi-mbta: It doesn't matter. Either, the process will stop abruptly. I was just trying to be consistent with the time.

supervisor/cubic_ods__process_incoming.py
grejdi-mbta: It depends on whether we want a graceful stop of a process or not. Most of the time you can get away with `while True:`, and not worry at all about it.

grejdi-mbta:
> * question: is `supervisor/` a Python module, or a directory for scripts? If it's the latter, I don't believe you need a `__init__.py`.
> 
>     * when I used `^C` to exit out of the container, it exited immediately: it didn't appear to wait for anything to shut down.
> 
> 
> Overall, this seems like a lot of boilerplate for something that's pretty straightforward in Elixir, and it would match our other ECS applications. Given that this application isn't interfacing directly with Spark (or even the data at all), does it make sense for it to be in Python?

@paulswartz 
* The supervisor process command is this `/usr/local/bin/python -m supervisor.cubic_ods__process_incoming`, so I added the `__init__.py` because of the `-m`
* I'm looking into the ^C to see if I missed anything.

Regarding your comment about Elixir. You're probably right. Elixir is more advanced in this sense. Although I have to be frank in that I would need ramp up on it to do this type of work. I mentioned this when we talked last week that this work can be done with Elixir, but I was choosing python so we can attract more talent in the data engineering area and make everything about the platform approachable by that talent. Maybe we can bring this up to a bigger group of folks, but we should make a decision soon, so I can continue to execute. Let me know what you think.

paulswartz:
Given that this application does not need to interface with Spark, I believe it would be better to have it written in Elixir. The data engineering is changing, with new libraries like [Polars](https://github.com/pola-rs/polars) in Rust and [Explorer](https://github.com/elixir-nx/explorer) in Elixir. It's also the language all of our other ECS tasks are in, meaning that we have lots of institutional knowledge about how to deploy/monitor/maintain Elixir services. 

If you still feel strongly that it should be in Python, an [RFC](https://github.com/mbta/technology-docs/tree/main/rfcs) would be the appropriate way to discuss with the larger engineering group.

PR #9, opened 2022-02-15
fix: make Credo stricter

paulswartz:
This should help the code be more Elixir-y overall, rather than needing
to rely on review comments.


grejdi-mbta (approved):
I'm excited and scared of this!

PR #10, opened 2022-02-16
fix: use correct paths for deps cache and Dialyzer

paulswartz:
`working-directory` at the top levle only affects `run` steps, not other
kinds of actions.


grejdi-mbta (approved):
@paulswartz Thanks.

PR #12, opened 2022-02-25
fix(ci): tell report-coverage where the code lives

paulswartz:


grejdi-mbta (approved):
Thanks @paulswartz. I took a stab at this in the beginning and needed to step back because I couldn't get it working.

PR #13, opened 2022-02-28
[elixir] starting glue job from oban for ingestion

grejdi-mbta:
For tracking: [Asana](https://app.asana.com/0/1201701089427502/1201865752934304/f)

This PR connects Oban to Glue, starting the glue jobs and monitoring for completion of them. Additionally, logic to group the loads into chunks for processing together is introduced. This will allow glue jobs to handle multiple load files in one job.

`ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex` has been updated to account for the preparation of the data that needs to be passed into the Oban job. Loads that don't have a table association are moved out to be processed as errors, and the rest are put into groups.

In `ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex`, the worker follows this process:
1. Get the list of load ids from job arguments.
2. Execute a request to run the Glue job for ingesting the loads.
3. Recursively get the Glue job's state to determine whether it is still running or not.
4. If not running anymore, then get the final state of the Glue job and update the status of the loads accordingly.


ex_cubic_ods_ingestion/.credo.exs
```diff
@@ -156,7 +156,7 @@
           {Credo.Check.Refactor.FilterReject, []},
           {Credo.Check.Refactor.IoPuts, []},
           {Credo.Check.Refactor.MapMap, []},
-          {Credo.Check.Refactor.ModuleDependencies, []},
+          {Credo.Check.Refactor.ModuleDependencies, [max_deps: 15]},
```
grejdi-mbta, line 159: Updated this, as I was hitting the default with 11 deps in the Oban worker module.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/schema/cubic_ods_load.ex
```diff
@@ -133,11 +133,6 @@ defmodule ExCubicOdsIngestion.Schema.CubicOdsLoad do
     Repo.all(query)
   end
 
-  @spec get!(integer()) :: t()
-  def get!(id) do
```
grejdi-mbta, line 137: Not needed anymore, so removed for now.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex
```diff
@@ -57,17 +58,32 @@ defmodule ExCubicOdsIngestion.StartIngestion do
...
+          {[{CubicOdsLoad.t(), nil}], [[{CubicOdsLoad.t(), CubicOdsTable.t()}]]}
+  def prepare_loads(load_recs) do
+    # attach table to load records, add or update any snapshots
+    loads = Enum.map(load_recs, &attach_table(&1))
+
+    {error_loads, ready_loads} =
+      Enum.split_with(loads, fn {_load_rec, table_rec} -> is_nil(table_rec) end)
+
+    # @todo replace chunk_every with chunk_while for more fine-tuned control
+    {error_loads, Enum.chunk_every(ready_loads, 3)}
```
grejdi-mbta, line 84: Creating a list of lists here, as I want to call the glue job with multiple loads.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex
```diff
@@ -104,18 +120,18 @@ defmodule ExCubicOdsIngestion.StartIngestion do
...
-    else
-      ProcessIngestion.error(load_rec)
-    end
-
-    :ok
+  @spec start_ingestion([{CubicOdsLoad.t(), CubicOdsTable.t()}]) :: {:ok, Oban.Job.t()}
+  def start_ingestion(loads_chunk) do
+    # update status to ingesting
+    load_recs_ids = Enum.map(loads_chunk, fn {load_rec, _table_rec} -> load_rec.id end)
+    CubicOdsLoad.update_many(load_recs_ids, status: "ingesting")
```
grejdi-mbta, line 127: Considering making this a transaction together with the Oban Job insert, but need to know more what `Ingest.new()` and `Oban.insert()` does.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,107 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
+  alias ExCubicOdsIngestion.ProcessIngestion
+  alias ExCubicOdsIngestion.Repo
 
-  require Logger
+  @impl Oban.Worker
+  def perform(%{args: args, meta: meta} = job) do
+    # get info passed into args
+    %{"chunk" => loads_chunk} = args
 
-  @job_timeout_in_sec 30
```
grejdi-mbta, line 14: Glue has its own timeout that we should rely on.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,107 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
-  @impl Oban.Worker
-  def perform(%Oban.Job{args: %{"load" => load, "table" => _table} = _args}) do
-    Process.sleep(2000)
+    # if no glue job run id available from meta data, then start the job
+    glue_job_run_id =
+      if is_nil(Map.get(meta, "glue_job_run_id")) do
+        {:ok, glue_job_run} =
+          start_glue_job_run(
+            lib_ex_aws,
+            "{\"a\": \"1\",\"b\": \"2\"}",
```
grejdi-mbta, line 33: Dummy data for now. Need to work on glue job a bit to figure out exactly what it needs.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,107 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
+            "{\"c\": \"3\",\"d\": \"4\"}"
+          )
+
+        # update job's meta data with glue job run id
```
grejdi-mbta, line 37: This ties the oban job to the glue job, in case we need to debug the process.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,107 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
+      end
 
-    load_rec = CubicOdsLoad.get!(load["id"])
-    CubicOdsLoad.update(load_rec, %{status: "ingested"})
+    # monitor for success or failure state in glue job run
+    case get_glue_job_run_status(lib_ex_aws, glue_job_run_id) do
+      {:ok, %{"JobRun" => %{"JobRunState" => "SUCCEEDED"}}} ->
+        ProcessIngestion.archive(loads_chunk)
+
+      _other_glue_job_run_state ->
```
grejdi-mbta, line 52: All other states are considered failures.

ex_cubic_ods_ingestion/test/ex_cubic_ods_ingestion/process_incoming_test.exs
```diff
@@ -41,43 +41,3 @@ defmodule ExCubicOdsIngestion.ProcessIncomingTest do
     # @todo test for a non-empty continuation token
   end
 end
-
-defmodule MockExAws do
```
grejdi-mbta, line 45: Moved to `test/support/`.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
paulswartz: Normally our process would be to log the Glue job ID: that way we can access it via Splunk without needing to dig directly into the RDS database.

ex_cubic_ods_ingestion/test/support/ex_aws.ex
```diff
@@ -0,0 +1,55 @@
+defmodule MockExAws do
+  @moduledoc """
+  MockExAws @todo
+  """
+
+  require Logger
+
+  @spec request(ExAws.Operation.t(), keyword) :: term
+  def request(op, _config_overrides \\ []) do
```
paulswartz, line 9: question: why have separate `request` and `request!` functions? You could simplify this to

```elixir
def request!(op, config_overrides) do
  case request(op, config_overrides) do
    {:ok, result} -> result
    {:error, error} -> raise ExAws.Error, inspect(error)
  end
end
```

(basically the original implementation that's down below)

ex_cubic_ods_ingestion/test/support/ex_aws.ex
```diff
@@ -0,0 +1,55 @@
...
+    cond do
+      op.service == :glue and Enum.member?(op.headers, {"x-amz-target", "AWSGlue.StartJobRun"}) ->
```
paulswartz, line 11: suggestion: this is better as a pattern match.
```elixir
def request(op, config_overrides \\ [])
def request(%{service: s3} = op, _config_overrides) do
  # ... S3 stuff ...
end
def request(%{service: :glue} = op, _config_overrides) do
  # ... glue stuff ...
end
def request(op, _config_overrides) do
  {:error, {:unknown_operation, op}}
end
```

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,107 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
+    glue_job_run = lib_ex_aws.request(operation)
+
+    case glue_job_run do
+      {:ok, %{"JobRun" => %{"JobRunState" => "RUNNING"}}} ->
+        get_glue_job_run_status(lib_ex_aws, run_id)
+
+      _other_glue_job_run_state ->
+        glue_job_run
+    end
+  end
```
paulswartz, line 112: suggestion: the AWS operation functionality seems like it could be pulled out into a module.
```elixir
defmodule ExAws.Glue do
  def start_job_run(job_name, argruments)
  def get_job_status(job_name, run_id)
end
```

And then in here:
```
lib_ex_aws.request(ExAws.Glue.start_glue_job(...))
lib_ex_aws.request(ExAws.Glue.get_job_status(...))
```

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,107 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
+          )
+
+        # update job's meta data with glue job run id
+        job
+        |> Changeset.change(meta: %{"glue_job_run_id" => glue_job_run["JobRunId"]})
+        |> Repo.update()
+
+        glue_job_run["JobRunId"]
+      else
+        Map.get(meta, "glue_job_run_id")
```
paulswartz, line 44: question: this will re-use the existing job name if this task restarts, correct? if this job crashes because of something about the Glue Job, will a new job ever be kicked off?

It might be better to store the job ID in `meta`, but not to use an existing ID if already present.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
paulswartz: It's probably worth keeping, even if we need to make it longer to incorporate the Glue job timeout. From Fred Hebert (https://erlang.org/pipermail/erlang-questions/2014-March/078175.html):

> To quote Pat Helland, if you're going to make
the default timeout infinity, I'd instead recommend setting it to 30
years. It's going to be nearly as long in practice, but will eventually
give up (that's guaranteed). It's not more ridiculous of an option, is
it?

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex
```diff
@@ -104,18 +120,18 @@ defmodule ExCubicOdsIngestion.StartIngestion do
...
+
+    # queue chunk for ingesting
+    %{
+      chunk:
+        Enum.map(loads_chunk, fn {load_rec, table_rec} -> %{load: load_rec, table: table_rec} end)
+    }
```
paulswartz, line 133: What about passing in a list of load record IDs? That would be easier to pass into other functions (`archive` and `error`), and avoids a lot of the complications with JSON serialization. It also ensures that the worker gets the latest version of any data, by re-fetching it from the DB when the job starts.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex
```diff
@@ -57,17 +58,32 @@ defmodule ExCubicOdsIngestion.StartIngestion do
...
-    # attach table to load records, and update snapshots
-    ready_loads = Enum.map(ready_load_recs, &attach_table(&1))
+    # prepare lists for separate operations
+    {error_loads, ready_loads_chunks} = prepare_loads(ready_load_recs)
 
-    # iterate through the load records list in order to start ingesting job,
-    # or error
-    Enum.each(ready_loads, &start_ingestion(&1))
+    # error out the error loads
+    Enum.each(error_loads, &ProcessIngestion.error(&1.id))
```
paulswartz, line 65: question: I thought `error` took a list now, or maybe a map? Either way, it seems like this isn't covered by tests. Maybe you could skip it for now with something like `[] = error_loads` (will crash if error loads is not an empty list).

ex_cubic_ods_ingestion/config/runtime.exs
```diff
@@ -10,7 +10,9 @@ config :ex_cubic_ods_ingestion,
   s3_bucket_incoming: System.get_env("S3_BUCKET_INCOMING"),
   s3_bucket_archive: System.get_env("S3_BUCKET_ARCHIVE"),
   s3_bucket_error: System.get_env("S3_BUCKET_ERROR"),
-  s3_bucket_springboard: System.get_env("S3_BUCKET_SPRINGBOARD")
+  s3_bucket_springboard: System.get_env("S3_BUCKET_SPRINGBOARD"),
+  glue_database: System.get_env("GLUE_DATABASE"),
```
paulswartz, line 14: question: is this used?

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_ingestion.ex
```diff
@@ -5,13 +5,15 @@ defmodule ExCubicOdsIngestion.ProcessIngestion do
 
   alias ExCubicOdsIngestion.Schema.CubicOdsLoad
 
-  @spec archive(CubicOdsLoad.t()) :: CubicOdsLoad.t()
-  def archive(load_rec) do
-    CubicOdsLoad.update(load_rec, %{status: "archived"})
+  @spec archive(map()) :: [integer()]
+  def archive(loads_chunk) do
+    load_recs_ids = Enum.map(loads_chunk, fn %{"load" => load_rec} = _chunk -> load_rec["id"] end)
```
paulswartz, line 10: nitpick: you don't need the ` = _chunk` bit. you could also do it as a single match: `fn %{"load" => %{"id" => id}} -> id end`

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/schema/cubic_ods_load.ex
```diff
@@ -149,6 +144,20 @@ defmodule ExCubicOdsIngestion.Schema.CubicOdsLoad do
     load_rec
   end
 
+  # @todo consider making this more specific to use cases
+  @spec update_many([t()], Keyword.t()) :: [integer()]
```
paulswartz, line 148: suggestion: this doesn't take a `[t()]`, it takes a `[integer()]`.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex
paulswartz: From the [Worker.new/2 documentation](https://hexdocs.pm/oban/Oban.Worker.html#c:new/2):
> Build a job changeset for this worker with optional overrides.

[Oban.insert/2](https://github.com/sorentwo/oban/blob/188cf9087547bc5f38b75ca8993575d064ad4e25/lib/oban/queue/basic_engine.ex#L53) uses a transaction, but could be combined with this update.

ex_cubic_ods_ingestion/.credo.exs
grejdi-mbta: No. I haven't cleaned up. Default is 10, so I was just over the limit. Will see how it looks after addressing code review issues.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,107 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
 
-  require Logger
+  @impl Oban.Worker
+  def perform(%{args: args, meta: meta} = job) do
+    # get info passed into args
+    %{"chunk" => loads_chunk} = args
 
-  @job_timeout_in_sec 30
+    # allow for ex_aws module to be passed in as a string, since Oban will need to
+    # serialize args to JSON. defaulted to library module.
```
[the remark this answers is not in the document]
grejdi-mbta: I mimicked `s3_import_worker.ex` from draft repo here. Did I miss something?

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_ingestion.ex
```diff
@@ -5,13 +5,15 @@ defmodule ExCubicOdsIngestion.ProcessIngestion do
 
   alias ExCubicOdsIngestion.Schema.CubicOdsLoad
 
-  @spec archive(CubicOdsLoad.t()) :: CubicOdsLoad.t()
-  def archive(load_rec) do
-    CubicOdsLoad.update(load_rec, %{status: "archived"})
+  @spec archive(map()) :: [integer()]
```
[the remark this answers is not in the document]
grejdi-mbta: > Where is this format coming from: could we provide this in a better format?

This is the payload for the oban job. It looks like this when it's created:
```
%{
      chunk:
        Enum.map(loads_chunk, fn {load_rec, table_rec} -> %{load: load_rec, table: table_rec} end)
    }
```

I can make it more specific, it just would be messy looking.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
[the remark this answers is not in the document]
paulswartz: `s3_import_worker.ex` doesn't have a loop in it: internal timeouts with `ExAws.request` will cause the worker to crash during processing if it takes too long. This worker could potentially loop indefinitely if the Glue job doesn't terminate.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex
grejdi-mbta: hmm.. ok. in general I don't like doing a lot within an async job. it introduces more failure points. I'll use IDs. I'm just a little apprehensive about it.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: Great quote!

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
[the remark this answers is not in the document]
grejdi-mbta: Noted on the timeout.
Regarding the security vulnerability, it seems to be in Draft as well. We should address there as well.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: This feels like 2 data points that should be connected, but I can move towards just logging it. Question for you, do you want to log it all the time or just in case of error states?

paulswartz: I would log the ID when the job is started, and then log the status after each check.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: I wanted to couple the Oban Job with the Glue Job. The uncertainty with having Oban kick off another Glue Job is worrisome. I prefer that the Job fails on both sides.

ex_cubic_ods_ingestion/test/support/ex_aws.ex
grejdi-mbta: Thanks, this is helpful. Still getting used to the language.

ex_cubic_ods_ingestion/config/config.exs
```diff
@@ -19,6 +19,14 @@ config :ex_cubic_ods_ingestion, Oban,
     ingest: 5
   ]
 
+config :ex_cubic_ods_ingestion, ExCubicOdsIngestion.Repo,
```
grejdi-mbta, line 22: Moved this here and out of `runtime.exs`, so I can make adjustments to it in `test.exs`.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,88 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
 
-  @job_timeout_in_sec 30
+  @log_prefix "[ex_cubic_ods_ingestion][workers][ingest]"
+  # 15 minutes
+  @job_timeout_in_sec 900
 
   @impl Oban.Worker
-  def perform(%Oban.Job{args: %{"load" => load, "table" => _table} = _args}) do
-    Process.sleep(2000)
+  def timeout(_job), do: :timer.seconds(@job_timeout_in_sec)
```
grejdi-mbta, line 19: @paulswartz Any tips on how to go about testing a job timeout? Was thinking of having the timeout be configurable and then checking the database.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: Logged now with `Logger.error`

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: Extracted. See `ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/aws/ex_aws_glue.ex`.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex
grejdi-mbta: Still looking into this. Mind if I add a `@todo` to come back to it.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_ingestion.ex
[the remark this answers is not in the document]
grejdi-mbta: Done.

ex_cubic_ods_ingestion/test/support/ex_aws.ex
grejdi-mbta: Done. And I'm now using `request!` throughout.

ex_cubic_ods_ingestion/test/support/ex_aws.ex
grejdi-mbta: Done.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: I removed the `meta` update. Regarding retry, I will remove retry on the glue job setting, and use Oban to retry. Let me know what you think.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/start_ingestion.ex
grejdi-mbta: Fixed. I had an issue because `run/1` was not being tested. Adjust and should be fully tested now.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/schema/cubic_ods_load.ex
grejdi-mbta: Fixed.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
paulswartz: I think this falls under "don't test the library". Oban's tests are responsible for ensuring that the timeout happens properly.

ex_cubic_ods_ingestion/.credo.exs
paulswartz: It looks like the error is coming from `start_ingestion.ex`, not `ingest.ex`. I think it's telling you that this module is doing too much, but we can come back to this particular issue later on.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,88 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
-    CubicOdsLoad.update(load_rec, %{status: "ingested"})
+        ProcessIngestion.error(load_rec_ids)
+    end
 
     :ok
   end
 
-  @impl Oban.Worker
-  def timeout(_job), do: :timer.seconds(@job_timeout_in_sec)
+  @spec start_glue_job_run(any(), String.t(), String.t()) :: map()
```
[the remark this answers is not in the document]
paulswartz: It's a `module()`: https://hexdocs.pm/elixir/1.13/typespecs.html#built-in-types describes all the built-in types.

ex_cubic_ods_ingestion/.credo.exs
grejdi-mbta: Addressed.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/aws/ex_aws_glue.ex
```diff
@@ -0,0 +1,39 @@
+defmodule ExAws.Glue do
```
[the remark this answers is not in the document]
grejdi-mbta: Done.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,88 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
+    )
+  end
+
+  @spec get_glue_job_run_status(any(), String.t()) :: map()
+  defp get_glue_job_run_status(lib_ex_aws, run_id) do
+    glue_job_name = Application.fetch_env!(:ex_cubic_ods_ingestion, :glue_job_cubic_ods_ingest)
+
+    # pause a litte before getting status
+    # Process.sleep(5000)
+    glue_job_run_status = lib_ex_aws.request!(ExAws.Glue.get_job_run(glue_job_name, run_id))
```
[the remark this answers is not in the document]
grejdi-mbta: Oversight. It was slowing down testing, so I commented out temporarily. It's back now.

ex_cubic_ods_ingestion/test/ex_cubic_ods_ingestion/workers/ingest_test.exs
```diff
@@ -0,0 +1,60 @@
...
+  setup do
+    # Explicitly get a connection before each test
+    # @todo check out https://github.com/mbta/draft/blob/main/test/support/data_case.ex
+    :ok = Sandbox.checkout(Repo)
+  end
+
+  describe "perform/1" do
+    test "enqueueing job" do
+      # @todo the assertion below is not working
+      # assert_enqueued(
```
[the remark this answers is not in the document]
grejdi-mbta: I didn't. From the documentation, it didn't seem that I needed to.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: Please let me know if you need more info here.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -7,22 +7,88 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
+
+    case glue_job_run_status do
+      %{"JobRun" => %{"JobRunState" => "RUNNING"}} ->
+        Logger.info("#{@log_prefix} Glue Job Run Status: #{Jason.encode!(glue_job_run_status)}")
```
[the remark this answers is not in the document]
grejdi-mbta: I log the non-running case out side of this function, line 51 and 56 above.

ex_cubic_ods_ingestion/test/ex_cubic_ods_ingestion/workers/ingest_test.exs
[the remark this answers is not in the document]
paulswartz: from [Oban.Testing.assert_enqueued/2](https://hexdocs.pm/oban/Oban.Testing.html#assert_enqueued/2):
> Assert that a job with particular options has been enqueued.

So if you didn't enqueue a job, this assertion would fail. If you expected a job to not be enqueued, you could use the reverse, `refute_enqueued/2`.

ex_cubic_ods_ingestion/lib/aws/ex_aws_glue.ex
```diff
@@ -0,0 +1,39 @@
+defmodule ExAws.Glue do
```
paulswartz, line 1: nitpick(non-blocking): should be `lib/ex_aws/glue.ex`

paulswartz (approved):
🍰

ex_cubic_ods_ingestion/lib/aws/ex_aws_glue.ex
grejdi-mbta: addressing in followup PR.

PR #14, opened 2022-02-28
feat: basic Dockerfile for ExCubicOdsIngestion

paulswartz:
Meets our Dockerfile best practices:

- multi-stage build
- `COPY` not `ADD`, and only the files which are needed
- separate non-root user account
- `dumb-init`

Will be needed once we are ready to deploy to ECS/Fargate.


grejdi-mbta:
@paulswartz this is good with me. I tested the Dockerfile. I'm going to bring into the main branch for some additional workflow work.

PR #18, opened 2022-03-05
[elixir+python] fix: add glue job info to ingest job

grejdi-mbta:
For reference: [Asana](https://app.asana.com/0/1201701089427502/1201894105231379/f)

In this PR, we update our Ingest worker to get more information for the load IDs it receives. We pass this information to the Glue Job, so it has everything it needs for a successful run.

In the Glue Job code, found at `aws/s3/glue_jobs/cubic_ods_ingest.py`, we make adjustments to be able to reference the input provided. We also adjust the Glue Job to overwrite the target. This helps with scenarios where the Glue Job might fail, and reattempted, by not appending (default) data.

Additionally, we clear out some run steps that are no longer needed during deployment. These run steps are now part of the more standard ECS container build that CTD uses throughout.


aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -27,51 +25,42 @@
...
-    },
-    format_options={ 'compression': 'gzip' },
-    transformation_ctx='{}_table_df_write_parquet'.format(jobName)
-  )
+  # convert to spark dataframe so we can use the withColumn functionality
+  tableSparkDF = tableDF.toDF()
+  # add partition columns
+  tableSparkDF = tableSparkDF.\
+                    withColumn('snapshot', lit(load.get('snapshot'))).\
+                    withColumn('identifier', lit(os.path.basename(load.get('s3_key'))))
```
grejdi-mbta, line 55: Helps with identifying what to overwrite.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -12,9 +12,7 @@
 from pyspark.sql.functions import *
 
 
-sc = SparkContext()
-glueContext = GlueContext(sc)
-spark = glueContext.spark_session
+glueContext = GlueContext(SparkContext())
```
grejdi-mbta, line 15: Working on figuring out a good testing strategy for this file.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -27,51 +25,42 @@
...
-  )
+  # convert to spark dataframe so we can use the withColumn functionality
+  tableSparkDF = tableDF.toDF()
+  # add partition columns
+  tableSparkDF = tableSparkDF.\
+                    withColumn('snapshot', lit(load.get('snapshot'))).\
+                    withColumn('identifier', lit(os.path.basename(load.get('s3_key'))))
+  # write out to springboard bucket with overwrite mode
+  tableSparkDF.write.\
+    mode('overwrite').\
```
paulswartz, line 58: issue: this is not the mode you want. it overwrites everything underneath the prefix you give it, so this would delete all the data from all snapshots on each load. You probably want `append`, and we'll handle any duplicate data on the processing end.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -62,13 +56,13 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
...
-  defp start_glue_job_run(lib_ex_aws, env_arg, input_arg) do
+  defp start_glue_job_run(lib_ex_aws, env_payload, input_payload) do
     glue_job_name = Application.fetch_env!(:ex_cubic_ods_ingestion, :glue_job_cubic_ods_ingest)
 
     lib_ex_aws.request!(
       ExAws.Glue.start_job_run(glue_job_name, %{
-        "--ENV": env_arg,
-        "--INPUT": input_arg
+        "--ENV": env_payload,
+        "--INPUT": input_payload
```
paulswartz, line 65: question: if we're providing both payloads, what's the difference between what goes into `ENV` and what goes into `INPUT`?

ex_cubic_ods_ingestion/test/ex_cubic_ods_ingestion/schema/cubic_ods_load_test.exs
```diff
@@ -139,4 +140,49 @@ defmodule ExCubicOdsIngestion.Schema.CubicOdsLoadTest do
...
+
+      assert [] == CubicOdsLoad.get_many_with_table(new_load_rec_ids)
+    end
+
+    test "getting records by passing load records with tables attached" do
+      # insert records as ready
+      {:ok, new_load_recs} = CubicOdsLoad.insert_new_from_objects(MockExAws.Data.load_objects())
+
+      # use the first record
+      first_load_rec = List.first(new_load_recs)
```
paulswartz, line 162: suggestion: rather than inserting alll the mock data only to get the first one, you could insert a single `%CubicOdsLoad{}` struct.

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: I default to Black, so will integrate that soon.

ex_cubic_ods_ingestion/test/ex_cubic_ods_ingestion/schema/cubic_ods_load_test.exs
grejdi-mbta: Yes. Laziness I guess, and it felt better to standardize insertion. Do you think it affects integrity of test? Happy to be more thorough.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: `ENV` is used to pass environment variables to Glue. Previously, I had a `.env` file that would be used by the dataplatform package, but it was proving hard to get everything in the package with CI. `INPUT` is used to pass data to the glue job. I separated them for clarity. No other reason. I will be documenting them.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -27,51 +25,42 @@
...
-    transformation_ctx='{}_table_df_write_parquet'.format(jobName)
-  )
+  # convert to spark dataframe so we can use the withColumn functionality
+  tableSparkDF = tableDF.toDF()
+  # add partition columns
+  tableSparkDF = tableSparkDF.\
+                    withColumn('snapshot', lit(load.get('snapshot'))).\
+                    withColumn('identifier', lit(os.path.basename(load.get('s3_key'))))
+  # write out to springboard bucket with overwrite mode
+  tableSparkDF.write.\
```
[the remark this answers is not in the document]
grejdi-mbta: Yes. I was using it before, but not sure if Spark accepts it as a format to its APIs. I will look into it.

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: I'd prefer to overwrite. Makes the system more tolerant to issues in the pipeline. The `identifier` will help with knowing exactly which area to overwrite. So if it's `LOAD1.csv`, we can replace just that chunk of the snapshot. Previously I was thinking that batch loads would just be appended, but if there is a problem with one and a retry is needed, we would have to clear out all the loads and retry all of them, instead of just one.

To address the other concerns about the using a partition for the `identifier`. I wanted use something to identify the prefix that I needed to overwrite. Initially, I was planning to just have the file name in the prefix after the `snapshot` partition, i.e. `cubic_ods_qlik/EDW.SAMPLE/snapshot=2022.../LOAD1.csv/`, but ultimately thought having an `identifier` partition might help with targeted sampling and granularity further down the pipeline. Let me know what you think.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -27,51 +25,42 @@
...
+    mode('overwrite').\
+    partitionBy('snapshot', 'identifier').\
```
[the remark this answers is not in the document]
grejdi-mbta: Commented about this here: https://github.com/mbta/data_platform/pull/18/files#r821878524

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: Are you working directly in the AWS Console? Or using the local Glue container.

aws/s3/glue_jobs/cubic_ods_ingest.py
paulswartz: Here's the script I used:
```python
from pyspark import SparkContext
from awsglue.context import GlueContext
from pyspark.sql.functions import *

glueContext = GlueContext(SparkContext.getOrCreate()) 
inputDF = glueContext.create_dynamic_frame_from_options(connection_type = "s3", connection_options = {"paths": ["s3://awsglue-datasets/examples/us-legislators/all/memberships.json"]}, format = "json")
df = inputDF.toDF()
df = df.withColumn('snapshot', lit("snap_id"))

df_one = df.withColumn('identifier', lit('one'))
df_one.show()

df_two = df.withColumn('identifier', lit('two'))
df_two.show()

df_one.write.mode('overwrite').partitionBy('snapshot', 'identifier').parquet('/tmp/memberships')
# simulate writing a different identifier
df_two.write.mode('overwrite').partitionBy('snapshot', 'identifier').parquet('/tmp/memberships')
```

And the output:
```bash
$ ls -R /tmp/memberships
/tmp/memberships/:                                                              
 _SUCCESS  'snapshot=snap_id'

'/tmp/memberships/snapshot=snap_id':
'identifier=two'

'/tmp/memberships/snapshot=snap_id/identifier=two':
part-00000-0dd7d8ec-755c-4c4d-9bb6-60e149570d8a.c000.snappy.parquet
```
Only the second write is kept: the first is completely overwritten even though the partition keys are different. Are you seeing something different if you run the script against a real Glue environment?

ex_cubic_ods_ingestion/test/ex_cubic_ods_ingestion/schema/cubic_ods_load_test.exs
paulswartz: I find it easier to understand (there's nothing particular about the data in `load_objects` which affects this test) and you don't need to pull out the first record. But it's a personal preference, so you don't need to take the suggestion :)

aws/s3/glue_jobs/cubic_ods_ingest.py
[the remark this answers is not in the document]
paulswartz: Resolving this to focus the conversation in the other thread.

aws/s3/glue_jobs/cubic_ods_ingest.py
paulswartz: See the other thread for the script I used.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -27,51 +25,42 @@
...
     dataCatalogTableName += '__ct'
+    loadTableS3Prefix = loadTableS3Prefix[:-1] + '__ct/'
 
   # create table dataframe using the data catalog table in glue
   tableDF = glueContext.create_dynamic_frame.from_catalog(
     database=envDict.get('GLUE_DATABASE_NAME'),
     table_name=dataCatalogTableName,
     additional_options={
-      'paths': table.get('load_s3_keys')
+      'paths': ['s3://{}/{}'.format(envDict.get('S3_BUCKET_INCOMING'), load.get('s3_key'))]
```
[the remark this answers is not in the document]
grejdi-mbta: Done.

aws/s3/glue_jobs/cubic_ods_ingest.py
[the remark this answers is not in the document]
grejdi-mbta: Looked into this and it seems to only be a format available when using dynamic frames. Because of the 'overwrite' mode, I'm unable to use dynamic frames, so I can't use the format.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -12,9 +12,7 @@
 from pyspark.sql.functions import *
 
 
-sc = SparkContext()
-glueContext = GlueContext(sc)
-spark = glueContext.spark_session
+glueContext = GlueContext(SparkContext())
 args = getResolvedOptions(sys.argv, ['JOB_NAME', 'ENV', 'INPUT'])
```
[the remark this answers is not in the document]
grejdi-mbta: Took a first stab at modularizing the job, and write some documentation. Please take a look, and let me know what you think.

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: I see the issue and have resolved with a Spark config update. See [this](https://github.com/TheCodeCache/learn-spark-concepts/blob/master/Dynamic%20Partition%20Overwrite.md) for more info.
Thanks for the keen eye!

Let me know what you think about `identifier` as a partition approach.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: Documented some on the Glue Job code. Please resolve if satisfactory.

ex_cubic_ods_ingestion/test/ex_cubic_ods_ingestion/schema/cubic_ods_load_test.exs
grejdi-mbta: I'll leave as is.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -12,66 +11,94 @@
...
+
+    # create job using the glue context
+    job = Job(glue_context)
+    # initialize job
+    job.init(job_name, args)
+
+    # run glue transformations for each cubic ods load
+    for load in input_dict.get("loads", []):
+        data_catalog_table_name = f'incoming__{load["table_name"]}'
+        load_table_s3_prefix = load.get("table_s3_prefix")
```
paulswartz, line 67: question: should this be `get("table_s3_prefix", "")` to default to the empty prefix?

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -12,66 +11,94 @@
...
+
+        # convert to spark dataframe so we can use the withColumn functionality
+        table_spark_df = table_df.toDF()
+        # add partition columns
+        table_spark_df = table_spark_df.withColumn(
+            "snapshot", lit(load.get("snapshot"))
+        ).withColumn("identifier", lit(os.path.basename(load_s3_key)))
+
+        # write out to springboard bucket with overwrite mode
+        spark.conf.set("spark.sql.sources.partitionOverwriteMode", "dynamic")
```
paulswartz, line 93: suggestion: this is a global configuration, so it can live outside the `for` loop.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -12,66 +11,94 @@
...
+    -------
+    dict, dict
+        Dictionaries containing environment variables and job data.
+
+
+    Examples
+    --------
+    >>> parse_job_arguments(
+    ...   '{"GLUE_DATABASE_NAME": "db"}',
+    ...   '{"loads":[{"s3_key":"s3/prefix/key","snapshot":"20220101","table_name":"table","table_s3_prefix":"s3/prefix/"}]}'
```
paulswartz, line 35: question: looking at this, it looks like `s3_key` and `table_s3_prefix` are duplicating information? Even if one is a `__ct`, I think you're adding the `__ct` back to the prefix anyways on lines 71-73.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -12,66 +11,94 @@
...
+        if os.path.dirname(load_s3_key).endswith("__ct"):
+            data_catalog_table_name += "__ct"
+            load_table_s3_prefix = f"{load_table_s3_prefix[:-1]}__ct/"
+
+        # create table dataframe using the data catalog table in glue
+        table_df = glue_context.create_dynamic_frame.from_catalog(
+            database=env_dict.get("GLUE_DATABASE_NAME"),
+            table_name=data_catalog_table_name,
+            additional_options={
+                "paths": [f's3://{env_dict.get("S3_BUCKET_INCOMING")}/{load_s3_key}']
```
paulswartz, line 80: question: if this takes multiple paths, could you pass in all of the incoming keys which are part of the same table?

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -12,66 +11,94 @@
...
+            },
+            transformation_ctx=f"{job_name}_table_df_read",
+        )
+
+        # convert to spark dataframe so we can use the withColumn functionality
+        table_spark_df = table_df.toDF()
+        # add partition columns
+        table_spark_df = table_spark_df.withColumn(
+            "snapshot", lit(load.get("snapshot"))
+        ).withColumn("identifier", lit(os.path.basename(load_s3_key)))
```
paulswartz, line 90: question: how will this work for batching, where we combine multiple sources (probably `__ct` but possible batch as well) into a single Parquet file? Or is the plan that we never batch, and it's always 1 input CSV -> 1 Parquet file?

paulswartz (commented):
There are still a lot of `.get()` calls here. I think you mentioned in an earlier comment that you prefer it, but it has pretty different semantics from `[]`. Can you say more about why you prefer `get()`?

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: Yes. That was the original approach, but I moved to a one-to-one approach because Spark will distribute work to various workers and the order they come back in might not be as originally passed in causing issues when writing them to parquet. This was not much of an issue for batch loads, but problematic for cdc loads. Lastly, being able to overwrite also validated the one-to-one approach.

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: Always one-to-one.

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: Refactored a bit, so it's addressed.

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: Addressed with a refactor.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -27,51 +25,42 @@
...
 job.init(jobName, args)
 
-# run glue transformations for each cubic ods table
-for table in inputDict.get('tables', []):
-  dataCatalogTableName = 'incoming__{}'.format(table.get('name'))
-  if table.get('load_is_cdc', False):
+# run glue transformations for each cubic ods load
+for load in inputDict.get('loads', []):
+  # determine the glue catalog table name
+  dataCatalogTableName = 'incoming__{}'.format(load.get('table_name'))
```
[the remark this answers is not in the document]
grejdi-mbta: Addressed.

grejdi-mbta:
@paulswartz Addressed the gets. Essentially, anything that is required will be use the `x["y"]` syntax. Anything that is not uses `x.get("y", "z")` and provides a default. My preference on `get()` is historic, and in general I like using one syntax to access keys in maps.

aws/s3/glue_jobs/cubic_ods_ingest.py
paulswartz: I think the distribution to workers with a different write order can also be true of a single large input file. I'm not sure the order the files are written in matters, but we can see how it works in practice. Were you able to test this with your AWS environment? I don't see any other kinds of testing for the Python code.

paulswartz (approved):
A couple non-blocking comments on the code, and a question about manual testing.

aws/s3/glue_jobs/cubic_ods_ingest.py
grejdi-mbta: They will be out of order when written to parquet. In the case of batch loads, it doesn't matter. For CDC loads, we need to be aware of this when working with them. CDC records can be reduced into 3 lists, inserts, updates, and deletes and executed on in that order. For inserts and deletes the order doesn't matter, but for updates it does. This is what makes think about the pipeline beyond the Springboard, and why I keep bringing it up.

aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -12,66 +11,103 @@
...
+    # initialize job
+    job.init(job_name, args)
+
+    # run glue transformations for each cubic ods load
+    for load in input_dict.get("loads", []):
+        load_s3_key = load["s3_key"]
+        load_table_s3_prefix = removeprefix(
+            f'{os.path.dirname(load_s3_key)}/',
+            env_dict.get("S3_BUCKET_PREFIX_INCOMING", "")
+        )
```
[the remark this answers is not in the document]
grejdi-mbta: Crazy, but it's only available in Python 3.9+, which Glue does not support yet.

aws/s3/glue_jobs/cubic_ods_ingest.py
paulswartz: The CDC records have a timestamp in them: the order they're written to Parquet shouldn't matter as long as we process them out of Parquet based on their timestamp.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: Added a `@todo` to move to function annotations.

PR #21, opened 2022-03-14
[infra] feat: deploy workflow + python ci

grejdi-mbta:


grejdi-mbta:
@paulswartz Any idea why I'm not seeing the actions in this PR over on the [Actions](https://github.com/mbta/data_platform/actions) tab?

paulswartz:
IIRC they show up when they're triggered on a branch, or when they're on the default branch. For deploys, your best bet is probably to have a PR which checks in empty versions of the files you want: we can merge that and then you can iterate on the contents in a subsequent PR.

PR #22, opened 2022-03-16
[infra+code] feat: add glue/python ci

grejdi-mbta:


aws/s3/glue_jobs/cubic_ods_ingest.py
```diff
@@ -1,114 +1,6 @@
...
-from awsglue.context import GlueContext
-from awsglue.job import Job
-from awsglue.dynamicframe import DynamicFrame
-from pyspark.sql.functions import *
 
 
-def removeprefix(text, prefix):
-    if text.startswith(prefix):
-        return text[len(prefix):]
+from py_cubic_ods_ingestion import ingest_incoming
```
grejdi-mbta, line 3: Really barebones here. Needs to be added to CI checks, but wasn't sure.

docker/Dockerfile.glue__local
```diff
@@ -34,4 +30,14 @@ ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
 ENV SPARK_HOME=/spark-2.4.3-bin-spark-2.4.3-bin-hadoop2.8
 RUN /glue/bin/gluesparksubmit --version
 
+# python packaging
+RUN pip install poetry
+# install py_cubic_ods_ingestion's required python dependencies
```
grejdi-mbta, line 35: The block below can be repeatable for additional data platform packages, i.e. `py_cubic_dmap_ingestion` or `py_cubic_datalake`.

py_cubic_ods_ingestion/lib/py_cubic_ods_ingestion/job_helpers.py
```diff
@@ -0,0 +1,50 @@
+import typing
```
grejdi-mbta, line 1: Not exact structure for testing, but wanted to show pytest running.

py_cubic_ods_ingestion/pyproject.toml
```diff
@@ -0,0 +1,51 @@
+# poetry
+[tool.poetry]
+name = "py_cubic_ods_ingestion"
+version = "0.1.0"
+description = ""
+authors = ["MBTA CTD"]
```
grejdi-mbta, line 6: @paulswartz any standard here for this?

py_cubic_ods_ingestion/pyproject.toml
```diff
@@ -0,0 +1,51 @@
+# poetry
```
grejdi-mbta, line 1: Opted for poetry here, per your suggestion. I usually worked with setuptools directly, but poetry seems good. It seems to be under heavy development. Looking at the repo with 1k+ issues and lots of PRs. I think our use case is not that complicated, so not worried. Just wondering if there will be breaking changes in the future. Need to keep an eye on it.

.github/workflows/ci.yaml
```diff
@@ -99,3 +99,49 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: asdf
+    defaults:
+      run:
+        working-directory: py_cubic_ods_ingestion
+
+    steps:
+      - name: Checkout repository code
+        uses: actions/checkout@v2
+      - uses: actions/checkout@v2
```
paulswartz, line 114:
```suggestion
```

docker/Dockerfile.glue__local
```diff
@@ -34,4 +30,14 @@ ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
...
+COPY ./py_cubic_ods_ingestion /data_platform/py_cubic_ods_ingestion
+WORKDIR /data_platform/py_cubic_ods_ingestion
+RUN poetry config virtualenvs.create false
+RUN poetry install --no-dev
+# install an editable link of py_cubic_ods_ingestion
+RUN pip install --editable .
```
paulswartz, line 41: question: I thought this happens automatically when you run `poetry install`?

py_cubic_ods_ingestion/lib/py_cubic_ods_ingestion/ingest_incoming.py
```diff
@@ -0,0 +1,73 @@
+import typing
```
paulswartz, line 1: suggestion(non-blocking): It would be great to get some linting in a future PR: I don't believe all this import is used.

py_cubic_ods_ingestion/pyproject.toml
paulswartz: You can put yourself and your MBTA e-mail address: I believe that's the format. If you don't want to do that, you can use `developer@mbta.com`.

py_cubic_ods_ingestion/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+name = "py_cubic_ods_ingestion"
+version = "0.1.0"
+description = ""
+authors = ["MBTA CTD"]
+packages = [
+  { include = "py_cubic_ods_ingestion", from = "lib" }
+]
+
+[tool.poetry.dependencies]
+python = "^3.7"
```
paulswartz, line 12: question: doesn't it depend on Glue and PySpark?

py_cubic_ods_ingestion/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+# mypy
+[tool.mypy]
+python_version = 3.7
+disallow_untyped_defs = true
+disallow_any_unimported = true
+no_implicit_optional = true
+check_untyped_defs = true
+warn_return_any = true
+warn_unused_ignores = true
+show_error_codes = true
```
paulswartz, line 38: question: some of these seems redundant: for example, do you need both `disallow_untyped_defs = true` and `check_untyped_defs = true`?

py_cubic_ods_ingestion/lib/py_cubic_ods_ingestion/job_helpers.py
paulswartz, line 1: question: I thought the Python [best practice](https://docs.python-guide.org/writing/structure/#the-actual-module) was to name the directory with the code after the module? So this would live under `py_cubic_ods_ingestion/py_cubic_ods_ingestion/job_helpers.py`.

docker/Dockerfile.glue__local
```diff
@@ -34,4 +30,14 @@ ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
...
+
 CMD ["echo", "Container 'glue__local' ready!"]
```
paulswartz, line 43: question: should these also get tested as a part of the "Docker" CI step?

.github/workflows/report-coverage.yaml
```diff
@@ -3,7 +3,7 @@ name: Report Elixir Coverage
 on:
   workflow_run:
     workflows:
-      - Elixir CI
```
paulswartz, line 6: suggestion(non-blocking): you could also update this to download/report on the `python-lcov` artifact.

paulswartz (commented):
question: can you say more about the `archive` directory that you moved the Lambdas to? If the intent is to keep them around for the future, they'll always be in the Git history so you don't need to put them in a particular directory.

.github/workflows/report-coverage.yaml
grejdi-mbta: I was going to, but thought to leave it for another PR.

docker/Dockerfile.glue__local
grejdi-mbta: That hasn't been my experience. There's lots of long threads about this on Github. I couldn't figure out a concrete answer.

docker/Dockerfile.glue__local
grejdi-mbta: Sorry, I don't understand your question. What's 'these'?

.github/workflows/report-coverage.yaml
grejdi-mbta: [Asana ticket](https://app.asana.com/0/1201701089427502/1201998951371772/f)

py_cubic_ods_ingestion/lib/py_cubic_ods_ingestion/ingest_incoming.py
grejdi-mbta: Yes. I'll add something.

py_cubic_ods_ingestion/lib/py_cubic_ods_ingestion/job_helpers.py
grejdi-mbta: I always disliked this structure. It felt like redundant information, since I was already in the repo/package. Bundling under `lib/` or `src/` made more sense. In this case, I mimicked the Elixir structure. Happy to change it the recommendation.

py_cubic_ods_ingestion/pyproject.toml
grejdi-mbta: Glue doesn't have an official package. Pyspark does, but it also is packaged in the Glue environment, see Dockerfile.glue__local.
For PySpark, I will be including it in the dev-dependencies, so I can do some testing with [this package](https://github.com/MrPowers/chispa).

py_cubic_ods_ingestion/pyproject.toml
grejdi-mbta: A lot of mypy configs catch each other. I removed `check_untyped_defs`.

grejdi-mbta:
@paulswartz regarding the 'archive' directory. I like to keep useful code around for fast access. Tracking it down in git, especially deletions is too slow and takes me out of my element. The 'archive' folder is a stop before getting outright deleted.

py_cubic_ods_ingestion/lib/py_cubic_ods_ingestion/job_helpers.py
paulswartz: Can you say more about how adding `lib` or `src` reduces the redundant information? You have `py_cubic_ods_ingestion` in the path twice either way.

docker/Dockerfile.glue__local
paulswartz: Apologies, I meant the two Dockerfiles which live in `docker/`. We have a CI step which ensures the ECS Dockerfile builds properly: should we have a similar check that the two other Dockerfiles also build properly?

py_cubic_ods_ingestion/lib/py_cubic_ods_ingestion/job_helpers.py
grejdi-mbta: Like I said, I mimicked the Elixir structure here. I would have `job_helpers.py`, etc. under `src/`, so it wouldn't duplicate.

docker/Dockerfile.glue__local
grejdi-mbta: These Dockerfiles will not get deployed to ECS or anything like that, so I don't necessarily see the benefit in validating that they build correctly every time we commit (or some other trigger). I guess if we are trying to guarantee that they are ready for use in local environments, then maybe we can. What do you think?

grejdi-mbta:
> > @paulswartz regarding the 'archive' directory. I like to keep useful code around for fast access. Tracking it down in git, especially deletions is too slow and takes me out of my element. The 'archive' folder is a stop before getting outright deleted.
> 
> If the `archive` folder is only for your use, do you need to check in it? You can have [local ignored paths](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file) and keep as much in the `archive` as is helpful.

Neat! Didn't know about this. I have removed them.

docker/Dockerfile.glue__local
paulswartz: It's up to you: I don't use them for local development :)

paulswartz (approved):
🍰

PR #24, opened 2022-03-17
[infra] feat: add deploy workflows

grejdi-mbta:
This PR updates the deploy Github Actions to allow for an actual deployment of Glue jobs, and elixir container for ingestion.


grejdi-mbta:
@paulswartz looks like ECS deployment is working now. I don't see anything in the workflow logs to indicate other wise. If you get a chance, please take a look ECS to see if the container is running. I will be doing a test of the pipeline a little later.

.editorconfig
```diff
@@ -1,18 +1,17 @@
 
 root = true
 
-[*.ex*]
+[*]
```
grejdi-mbta, line 4: Fixed order of operations here.

.github/workflows/docker.yml
```diff
@@ -1,12 +0,0 @@
-name: ExCubicOdsIngestion Docker build
```
paulswartz, line 1: question: can we keep this? We've historically had issues where the Docker container doesn't build, which would have been caught if CI tried to build the container.

.github/workflows/glue-python-deploy.yaml
```diff
@@ -21,3 +21,18 @@ jobs:
     steps:
       - name: Checkout repository code
         uses: actions/checkout@v2
+      - name: Setup Python 3.7
```
paulswartz, line 24: question: should this be using `ASDF` like the main CI step?

.github/workflows/glue-python-deploy.yaml
```diff
@@ -21,3 +21,18 @@ jobs:
...
+        working-directory: ${{github.workspace}}
+        run: pip install -r requirements.txt -t ./libs
+      - name: Create archive, and add 'data_platform' folder
+        run: zip -r ${{github.workspace}}/aws/s3/packages/data_platform.zip ./data_platform/
+      # change directory 'libs', so libraries can be loaded into the root of archive
+      - name: Add python libs to archive
+        working-directory: ${{github.workspace}}/libs
+        run: zip -r ${{github.workspace}}/aws/s3/packages/data_platform.zip ./
+      - name: Sync Glue Jobs and package to operations
+        run: aws s3 sync ${{github.workspace}}/aws/s3/ s3://mbta-ctd-dataplatform-${{github.event.inputs.env}}-operations/ --exclude "*/empty"
```
paulswartz, line 38: question(non-blocking): this will work for the `dev` environment, but in production won't the bucket be `mbta-ctd-dataplatform-operatations`?

grejdi-mbta: Yes. I'm aware of that. I will fix once we've deployed a prod resources.

.github/workflows/glue-python-deploy.yaml
```diff
@@ -21,3 +21,18 @@ jobs:
     steps:
       - name: Checkout repository code
         uses: actions/checkout@v2
+      - name: Setup Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+      - name: Install python packages in 'libs' folder
```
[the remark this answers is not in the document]
grejdi-mbta: I had the code sitting in another branch until the CI got merged into 'main'. It's in now.

.github/workflows/glue-python-deploy.yaml
grejdi-mbta: It's in now.

.github/workflows/docker.yml
grejdi-mbta: I re-added. I just made the step more specific, so we can add more docker builds later on..

paulswartz (approved):
:cake:

PR #25, opened 2022-03-23
[code] fix: tests were not working with prefixes env variables set

grejdi-mbta:
This PR addresses an issue in testing with bucket prefix environment variables being set, specifically the `S3_BUCKET_PREFIX_INCOMING`.

An [Asana ticket](https://app.asana.com/0/1201701089427502/1201998951371780/f) has also been created for an ADR to explain bucket prefixes and their purpose.

### DIFF
--- a/ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_incoming.ex
+++ b/ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_incoming.ex
@@ -75,6 +75,9 @@ defmodule ExCubicOdsIngestion.ProcessIncoming do
         lib_ex_aws: state.lib_ex_aws
       )
       |> Enum.filter(&Map.has_key?(&1, :key))
+      |> Enum.map(fn object ->
+        %{object | key: String.replace_prefix(object[:key], incoming_prefix, "")}
+      end)
       |> CubicOdsLoad.insert_new_from_objects_with_table(table)
     end
 
--- a/ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_incoming.ex
+++ b/ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_incoming.ex
@@ -54,13 +54,13 @@ defmodule ExCubicOdsIngestion.ProcessIncoming do
   # server helper functions
   @spec run(t) :: :ok
   def run(state) do
-    bucket = Application.fetch_env!(:ex_cubic_ods_ingestion, :s3_bucket_incoming)
-    prefix = Application.fetch_env!(:ex_cubic_ods_ingestion, :s3_bucket_prefix_incoming)
+    incoming_bucket = Application.fetch_env!(:ex_cubic_ods_ingestion, :s3_bucket_incoming)
+    incoming_prefix = Application.fetch_env!(:ex_cubic_ods_ingestion, :s3_bucket_prefix_incoming)
 
     table_prefixes =
-      bucket
+      incoming_bucket
       |> S3Scan.list_objects_v2(
-        prefix: "#{prefix}cubic_ods_qlik/",
+        prefix: "#{incoming_prefix}cubic_ods_qlik/",
         delimiter: "/",
         lib_ex_aws: state.lib_ex_aws
       )
@@ -69,8 +69,11 @@ defmodule ExCubicOdsIngestion.ProcessIncoming do
       |> CubicOdsTable.filter_to_existing_prefixes()
 
     for {table_prefix, table} <- table_prefixes do
-      bucket
-      |> S3Scan.list_objects_v2(prefix: table_prefix, lib_ex_aws: state.lib_ex_aws)
+      incoming_bucket
+      |> S3Scan.list_objects_v2(
+        prefix: "#{incoming_prefix}#{table_prefix}",
+        lib_ex_aws: state.lib_ex_aws
+      )
       |> Enum.filter(&Map.has_key?(&1, :key))
       |> CubicOdsLoad.insert_new_from_objects_with_table(table)
     end
--- a/ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_incoming.ex
+++ b/ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_incoming.ex
@@ -65,7 +65,7 @@ defmodule ExCubicOdsIngestion.ProcessIncoming do
         lib_ex_aws: state.lib_ex_aws
       )
       |> Stream.filter(&Map.has_key?(&1, :prefix))
-      |> Enum.map(&Map.fetch!(&1, :prefix))
+      |> Enum.map(fn %{prefix: prefix} -> String.replace_prefix(prefix, incoming_prefix, "") end)
       |> CubicOdsTable.filter_to_existing_prefixes()
 
     for {table_prefix, table} <- table_prefixes do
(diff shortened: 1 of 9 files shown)


ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/schema/cubic_ods_table.ex
```diff
@@ -54,10 +54,15 @@ defmodule ExCubicOdsIngestion.Schema.CubicOdsTable do
   """
   @spec filter_to_existing_prefixes(Enumerable.t()) :: [{String.t(), t()}]
   def filter_to_existing_prefixes(prefixes) do
-    # strip any change tracking suffix
+    incoming_prefix = Application.fetch_env!(:ex_cubic_ods_ingestion, :s3_bucket_prefix_incoming)
+
+    # strip incoming bucket prefix as it will not be in the database
+    without_incoming_prefix = Enum.map(prefixes, &String.replace_prefix(&1, incoming_prefix, ""))
```
paulswartz, line 60: question: why shouldn't this prefix be in the database? If the data lives under "/incoming/cubic_ods_qlik/EDW.TABLE". shouldn't we configure that as the `s3_prefix`?

grejdi-mbta: I'd like the table prefix to be consistent across all transitions/buckets. If we store it, then when saving it to the Springboard bucket, the key will be `s3://{springboarg_bucket}/{springboard_prefix}/incoming/cubic_ods_qlik/EDW.TABLE/...`. Additionally, I'm looking at bucket prefixes as being more useful for organizing rather that part of the platform internals.

paulswartz: Do you need to use the incoming prefix when writing to the springboard? You already have the table name in `ExCubicOdsIngestion.CubicOdsTable.name` so you can write to `s3://<springboard bucket>/<springboard prefix>/cubic_ods_qlik/<name>/...`

grejdi-mbta: Actually, I misspoke. I don't use the table's s3_prefix when storing to the springboard. It's easier to just get it from the load object's key. 
```py
load_table_s3_prefix = removeprefix(
    f'{os.path.dirname(load_s3_key)}/',
    env_dict.get("S3_BUCKET_PREFIX_INCOMING", "")
)
```
By storing it without the 'incoming/' in the database though, I make sure that the table's s3_prefix represents a path that is valid whether you are in springboard, incoming, archive, or error bucket/prefix combos.

In live and dev environments, the bucket prefixes will be empty because we have different buckets for each stage. You might be arguing why add this complexity at all. The reason for me is simple. I'd like to be able to test the system from a local environment, hence the 'dataplatform-local' bucket. We could create different 'local' buckets to mimic the dev/prod environments, but if we scale beyond 1 engineer they might trip over each other. With prefixes, I'm able to create my own space without the need for a multiple buckets. Other data engineers will be able to do the same.
```
S3_BUCKET_PREFIX_OPERATIONS=gg/operations/
S3_BUCKET_PREFIX_INCOMING=gg/incoming/
S3_BUCKET_PREFIX_ARCHIVE=gg/archive/
S3_BUCKET_PREFIX_ERROR=gg/error/
S3_BUCKET_PREFIX_SPRINGBOARD=gg/springboard/
```

paulswartz: I get the reason for the prefixes: that makes sense to me. But with the individual developers having unique databases, I'm not sure why the prefixes can't be stored as a part of the S3 keys? And aren't the paths different in `springboard`,`archive`, and `error` anyways, because of the partitioning? 

Maybe an example would help me see what you're thinking.

paulswartz: To be honest, I'm still confused about what the data will look like :) Even if people have the same table configuration, presumably the CubicOdsLoad records will be different, as will the Glue Catalog Tables and the data in S3.

I think this would be clearer for me at least if the prefix stripping happens outside of this function, which otherwise doesn't care about the prefixes. You're calling it from `ProcessIncoming` which is already handling the `incoming_prefix`: could they be stripped there instead?

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_incoming.ex
```diff
@@ -54,23 +54,26 @@ defmodule ExCubicOdsIngestion.ProcessIncoming do
...
     for {table_prefix, table} <- table_prefixes do
-      bucket
-      |> S3Scan.list_objects_v2(prefix: table_prefix, lib_ex_aws: state.lib_ex_aws)
+      incoming_bucket
+      |> S3Scan.list_objects_v2(
+        prefix: "#{incoming_prefix}#{table_prefix}",
+        lib_ex_aws: state.lib_ex_aws
+      )
       |> Enum.filter(&Map.has_key?(&1, :key))
       |> CubicOdsLoad.insert_new_from_objects_with_table(table)
```
paulswartz, line 78: Do you need to strip the `incoming_prefix` out from the new objects as well?

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/schema/cubic_ods_table.ex
grejdi-mbta: I have moved it out.

I plan on writing a more detailed doc, but wanted to explain my intentions. My goal is to achieve testability at 2 levels out of the box, with unit testing and with actual infrastructure. What I mean with 'out of the box' is that you can download the code, run the migrations, adjust configs for local development (mainly the prefixes), and run. At this point and with the understanding of how the data platform's architecture works, you should be able to take a file from live and upload it your 'incoming' area. The elixir app will take it, run the 'dev' glue job, and send the new file to your 'springboard' area, as well as send the original file to you 'archive' area. For all of this to work, you need the 'dev' infrastructure up, and a local database that is the same as the 'dev'/'live' in its schema and the data for the CubicOdsTable, which is needed for recognizing files in the 'incoming' area. You'll notice I haven't mentioned CubicOdsLoad data. That's because those are not referenced when initially running the platform. They are only referenced after they are in the system, turning it more into just a log.

paulswartz: I get, and agree with, the general idea. I think a document, which could include some more specific examples and values, would be helpful: both for me right now, and for folks working with this repository in the future.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/process_incoming.ex
grejdi-mbta: Yes. Was working on that. Done now.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/schema/cubic_ods_table.ex
grejdi-mbta: I have this [Asana ticket](https://app.asana.com/0/1201701089427502/1201998951371780/f) and will tackle it next week. I was waiting for the container to be up and running to see if I needed to make any adjustment in the approach.

paulswartz (approved):
🍰 Excited to see that documentation in the future!

PR #30, opened 2022-04-05
[code] fix: snapshot now in iso 8601 + fix test

grejdi-mbta:
This PR adjust the format of the snapshot string to be ISO 8601 compatible. It also fixes a test that didn't have the proper data setup.

### DIFF
--- a/ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/archive.ex
+++ b/ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/archive.ex
@@ -37,7 +37,7 @@ defmodule ExCubicOdsIngestion.Workers.Archive do
       Enum.join([
         archive_prefix,
         Path.dirname(load_rec.s3_key),
-        "/snapshot=#{Calendar.strftime(load_rec.snapshot, "%Y%m%d%H%M%S")}/",
+        "/snapshot=#{Calendar.strftime(load_rec.snapshot, "%Y%m%dT%H%M%SZ")}/",
         Path.basename(load_rec.s3_key)
       ])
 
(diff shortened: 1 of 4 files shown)


ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/archive.ex
```diff
@@ -37,7 +37,7 @@ defmodule ExCubicOdsIngestion.Workers.Archive do
       Enum.join([
         archive_prefix,
         Path.dirname(load_rec.s3_key),
-        "/snapshot=#{Calendar.strftime(load_rec.snapshot, "%Y%m%d%H%M%S")}/",
+        "/snapshot=#{Calendar.strftime(load_rec.snapshot, "%Y%m%dT%H%M%SZ")}/",
```
paulswartz, line 40: question: are the snapshot times in UTC? You can use `"%z"` to get the offset in minutes.
```elixir
iex> Calendar.strftime(DateTime.utc_now(), "%Y%m%d%T%H%M%S%z")
"20220405T195026+0000"
```

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -110,7 +110,7 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
       Enum.map(CubicOdsLoad.get_many_with_table(load_rec_ids), fn {load_rec, table_rec} ->
         %{
           s3_key: load_rec.s3_key,
-          snapshot: load_rec.snapshot,
+          snapshot: Calendar.strftime(load_rec.snapshot, "%Y%m%dT%H%M%SZ"),
```
paulswartz, line 113: question: do you need to convert the snapshot time at this point?

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/archive.ex
grejdi-mbta: `+` means something in the URL, so I rather not use it. The way it is right now conforms as the timestamps are UTC.

ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
grejdi-mbta: I could do it in the glue job with python. But it's going to be converted to string with `Jason.encode/1`, so just got ahead of it.

paulswartz (approved):
suggestion(non-blocking): you could pull the ISO conversion into a function that all of these use, but that could also happen later.

PR #33, opened 2022-04-12
[code] fix: wrong table is referenced for change tracking load

grejdi-mbta:
This PR addresses an issue where the correct glue table was not being referenced for change tracking loads.


paulswartz (approved):
🍰 

suggestion(non-blocking): it might be valuable to pull out all of the code for calculating the `from_catalog` arguments into a function you can test. That would have caught the original issue as well. Something like:
```python
for load in input_dict.get("loads", []):
  from_catalog_kwargs = job_helpers.from_catalog_kwargs(load, env_dict)
  table_df = glue_context.create_dynamic_frame.from_catalog(**from_catalog_kwargs)
  ...
```

PR #46, opened 2022-05-18
[elixir] fix: glue throttling because of aggressiveness on oban's part.

grejdi-mbta:
After doing a load test on the ingestion process, I ran into an issue with throttling because of Glue's job concurrency limits. This throttling and Oban's speed created many other errors in the process as Oban moved to execute on other jobs in its pipeline.
Initially, I was hoping to be make a determination on when exactly Glue was actually done with a job, so I can start the next one, but upon further investigation Glue does not actually report this in its `Glue.get_job_run()` which contains the latest status of the job. So ultimately, I resulted in using a "low-tech" approach of just giving AWS time to sync its status across systems.

### DIFF
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex
@@ -35,53 +35,22 @@ defmodule ExCubicIngestion.Workers.Ingest do
       end
 
     # start glue job
-    {env_payload, input_payload} = construct_glue_job_payload(load_rec_ids)
-    glue_job_start_request = start_glue_job_run(lib_ex_aws, env_payload, input_payload)
+    glue_job_start_request =
+      load_rec_ids
+      |> construct_glue_job_payload()
+      |> start_glue_job_run(lib_ex_aws)
 
     case glue_job_start_request do
       {:ok, %{"JobRunId" => glue_job_run_id}} ->
-        Logger.info("#{@log_prefix} Glue Job Run ID: #{glue_job_run_id}")
+        monitor_glue_job_run(glue_job_run_id, load_rec_ids, lib_ex_aws)
 
-        # monitor for success or failure state in glue job run
-        glue_job_run_status = get_glue_job_run_status(lib_ex_aws, glue_job_run_id)
-
-        case glue_job_run_status do
-          %{"JobRun" => %{"JobRunState" => "SUCCEEDED"}} ->
-            Logger.info(
-              "#{@log_prefix} Glue Job Run Status: #{Jason.encode!(glue_job_run_status)}"
-            )
-
-            CubicLoad.update_many(load_rec_ids, status: "ready_for_archiving")
-
-            :ok
-
-          _other_glue_job_run_state ->
-            # note: error will be handled within ObanIngestWorkerError module
-            {:error, "Glue Job Run Status: #{Jason.encode!(glue_job_run_status)}"}
-        end
-
-      {:error, {"ConcurrentRunsExceededException", message}} ->
-        Logger.warning(
-          "#{@log_prefix} Glue Job Start Request: ConcurrentRunsExceededException: #{message}"
-        )
-
-        {:snooze, 60}
-
-      {:error, {"ThrottlingException", message}} ->
-        Logger.warning("#{@log_prefix} Glue Job Start Request: ThrottlingException: #{message}")
-
-        {:snooze, 60}
-
-      # everything else is an error
       _glue_job_start_request_error ->
-        Logger.error("#{@log_prefix} Glue Job Start Request: #{glue_job_start_request}")
-
-        {:error, "Glue Job Start Request: #{glue_job_start_request}"}
+        handle_start_glue_job_error(glue_job_start_request)
     end
   end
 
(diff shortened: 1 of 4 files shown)


ex_cubic_ods_ingestion/lib/ex_cubic_ods_ingestion/workers/ingest.ex
```diff
@@ -81,7 +85,27 @@ defmodule ExCubicOdsIngestion.Workers.Ingest do
 
     # pause a litte before getting status
     Process.sleep(5000)
-    glue_job_run_status = lib_ex_aws.request!(ExAws.Glue.get_job_run(glue_job_name, run_id))
+
+    glue_job_run_status =
```
grejdi-mbta, line 89: Updated this upon realizing that an error to get status does not necessarily mean that the job has errored. Looking for some feedback on the approach, especially the catch all.

ex_cubic_ods_ingestion/test/ex_cubic_ods_ingestion/workers/ingest_test.exs
```diff
@@ -9,6 +9,8 @@ defmodule ExCubicOdsIngestion.Workers.IngestTest do
   require MockExAws
 
   describe "perform/1" do
+    @tag timeout: 300_000
```
grejdi-mbta, line 12: Tests have a default timeout limit of 60 seconds. This overrides the default.

grejdi-mbta:
@paulswartz This was tested with 200+ change tracking loads, and it performed without a hitch. I will be running a bigger test, once the refactor PRs are approved and merged in.

paulswartz:
Can you say more about the throttling issue you ran into? I'm having trouble understanding how waiting for 60s after Glue completes is addressing the problem.

grejdi-mbta:
> Can you say more about the throttling issue you ran into? I'm having trouble understanding how waiting for 60s after Glue completes is addressing the problem.

Upon getting a "SUCCEEDED" status from the glue job, Oban completed its job, and then quickly spun another job to kickstart another glue job. Glue however doesn't seem to consider "SUCCEEDED" as the job is done. There are some [other states](https://docs.aws.amazon.com/glue/latest/dg/job-run-statuses.html) that apparently it goes through. Initially I thought I can tap into the "STOPPING" and/or "STOPPED" states, but glue doesn't report them when querying for status. So I opted for waiting, as I suspect that Glue needs a little bit of time to either stop the job and/or sync its state so it considers the job done.

paulswartz:
And you were seeing "ThrottlingException" errors? Those should be handled by `ExAws.request` with an exponential backoff, so I'm a little confused as to how they made it back to you. I also don't see that exception in the `dataplatform-dev` logs. Do you have any logs from the local run you could share?

grejdi-mbta:
> And you were seeing "ThrottlingException" errors? Those should be handled by `ExAws.request` with an exponential backoff, so I'm a little confused as to how they made it back to you. I also don't see that exception in the `dataplatform-dev` logs. Do you have any logs from the local run you could share?

Sharing the logs from local. You'll see a lot of `ConcurrentRunsExceededException`s and then a few `ThrottlingException`s.  I won't know I have exceeded concurrent runs until I run another job, but I rather not do that and just give space to Glue to finish. That's why I'm using the `Process.sleep/0`. I guess we could argue that we should just catch the `ConcurrentRunsExceededException/ThrottlingException` when we run another Glue job and just keep checking when it's available, but it feels wrong to wait then, rather than at the end of a current job. Let me know if you're thinking something else.

[local_logs.txt](https://github.com/mbta/data_platform/files/8730142/local_logs.txt)

paulswartz:
To me, it makes more sense to wait at the beginning of the new job. From the old Oban job's perspective, it's done with the work it set out to do. If for some reason the job crashed after seeing the Glue job SUCCEEDED status, there wouldn't be a need to restart the Oban job even if the 60s timeout was up.

The new job can wait a few seconds and retry, or possibly use Oban's built-in ["snooze" logic](https://hexdocs.pm/oban/Oban.Worker.html#module-snoozing-jobs). This also has the advantage of being easier to test, as you can check whether `perform/1` correctly returns `{:snooze, _}` without having to actually wait in the test case.

grejdi-mbta:
@paulswartz This is not ready, but wanted to note a couple of observations in case you have any feedback.
* Using the `:snooze` feature works well and I was able to run without error with about 100+ load files coming through. I feel confident that it will scale beyond that.
* Once a job snoozes, the next job starts which hits the same issue of `ConcurrentRunsExceededException`, and so it goes for all of the rest of the jobs queued up, and essentially snoozing all of them. We might be able to slow this down with a `Process.sleep(5_000)` or speed it up with a `{:snooze, 5}`, both achieving the same result at the expense of Glue's API hits.
* Currently, logging these `ConcurrentRunsExceededException` as `warning`s, but because they are expected I almost feel like just downgrading them to `info`s.

paulswartz:
That all seems fine, and I agree on reducing the level for ConcurrentRunsExceededException as long as we're expecting it as an error.

grejdi-mbta:
@paulswartz Was able to test this with double the concurrency value for Glue, and it eliminated `ConcurrentRunsExceededException`s as expected. The snooze logic is still in here as a fall back.

paulswartz (approved):
🍰

PR #48, opened 2022-05-20
[db] feat: add more ODS table records

grejdi-mbta:
This PR adds a migration for adding the rest of the ODS tables to the ingestion process.


grejdi-mbta:
@paulswartz I will not be deploying this until there is a successful run through with just `USE_TRANSACTION` and the new refactored code. Just wanted you to give it thumbs up.

paulswartz (approved):
🍰

PR #49, opened 2022-05-26
[elixir] fetch dmap worker

grejdi-mbta:


grejdi-mbta:
@paulswartz Feel free to peek in. Still working through the upload to S3 and some small Credo refactors, but overall should be in good shape.

.env.template
```diff
@@ -29,3 +29,8 @@ S3_BUCKET_PREFIX_SPRINGBOARD=
 GLUE_DATABASE_INCOMING=
 GLUE_DATABASE_SPRINGBOARD=
 GLUE_JOB_CUBIC_INGESTION_INGEST_INCOMING=
+
+# dmap
+DMAP_BASE_URL=
+DMAP_DATASET_PUBLIC_USERS_API_KEY=
+DMAP_CONTROLLED_RESEARCH_USERS_API_KEY=
```
paulswartz, line 36: suggestion: you only need one of these keys. a CR API key can be used for both sets of API calls.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_dmap_dataset.ex
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  defp upsert_from_dataset(dataset, feed_rec) do
+    rec = get_by(identifier: dataset["dataset_id"])
+
+    if rec do
+      Repo.update!(
+        Changeset.change(rec, %{
+          start_date: Date.from_iso8601!(dataset["start_date"]),
+          end_date: Date.from_iso8601!(dataset["end_date"]),
```
paulswartz, line 99: I don't think it's possible for these values to change on a given dataset.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_dmap_dataset.ex
```diff
@@ -0,0 +1,120 @@
...
+        type: dataset["id"],
+        identifier: dataset["dataset_id"],
+        start_date: Date.from_iso8601!(dataset["start_date"]),
+        end_date: Date.from_iso8601!(dataset["end_date"]),
+        last_updated_at:
+          dataset["last_updated"]
+          |> Timex.parse!("{ISO:Extended}")
+          |> DateTime.from_naive!("Etc/UTC")
+      })
+    end
```
paulswartz, line 118: suggestion: there might be a way to do an actual upsert. Check in `mbta/skate` because I think they use it for notifications.

ex_cubic_ingestion/lib/ex_cubic_ingestion/utility.ex
```diff
@@ -0,0 +1,26 @@
+defmodule ExCubicIngestion.Utility do
```
paulswartz, line 1: suggestion: given the contents of this file, maybe `ExCubicIngestion.Validators` would be a more specific name?

ex_cubic_ingestion/lib/ex_cubic_ingestion/utility.ex
```diff
@@ -0,0 +1,26 @@
...
+  @moduledoc """
+  Module for holding helpful functions
+  """
+
+  @spec is_valid_date(String.t()) :: boolean()
+  def is_valid_date(date_str) do
```
paulswartz, line 7: suggestion: since these can't be used as guards, the naming convention would be `valid_date?`

ex_cubic_ingestion/lib/ex_cubic_ingestion/utility.ex
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  @spec is_valid_url(String.t()) :: boolean()
+  def is_valid_url(url) do
+    URI.parse(url).scheme == "https" &&
+      URI.parse(url).path not in [nil, "/"]
+  end
+
+  @spec map_has_keys(map(), [String.t()]) :: boolean()
+  def map_has_keys(map, key_list) do
```
paulswartz, line 23: suggestion: `map_has_keys?`

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/fetch_dmap.ex
```diff
@@ -0,0 +1,136 @@
...
+        _args_lib_httpoison -> HTTPoison
+      end
+
+    feed_rec = CubicDmapFeed.get!(feed_id)
+
+    feed_rec
+    |> construct_feed_url(last_updated)
+    |> get_feed(lib_httpoison)
+    |> Map.get("results", [])
+    |> Enum.filter(&is_valid_dataset(&1))
```
paulswartz, line 47: suggestion: `valid_dataset?`

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/fetch_dmap.ex
```diff
@@ -0,0 +1,136 @@
...
+    dmap_base_url = Application.fetch_env!(:ex_cubic_ingestion, :dmap_base_url)
+
+    dmap_dataset_public_users_api_key =
+      Application.fetch_env!(:ex_cubic_ingestion, :dmap_dataset_public_users_api_key)
+
+    dmap_controlled_research_users_api_key =
+      Application.fetch_env!(:ex_cubic_ingestion, :dmap_controlled_research_users_api_key)
+
+    api_key =
+      if String.starts_with?(feed_rec.relative_url, "/controlledresearchusersapi") do
```
paulswartz, line 70: suggestion: I see what you're doing here, but all of the feeds are available under this endpoint, even the aggregate ones. You can configure all the relative URLs to be under `/controlledresearcherusersapi` and then only have one API key.

paulswartz (commented):
should the sample files have data?

ex_cubic_ingestion/lib/ex_cubic_ingestion/utility.ex
grejdi-mbta: Done.

ex_cubic_ingestion/lib/ex_cubic_ingestion/utility.ex
grejdi-mbta: Done. Gotta get used to adding `?` for these types of functions.

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/fetch_dmap.ex
grejdi-mbta: Addressed.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_dmap_dataset.ex
paulswartz: It was users/user settings, not notifications: my mistake!

- https://github.com/mbta/skate/blob/606cf30acf93aa656029b4dd5cd676c17116f5c7/lib/skate/settings/user.ex#L6-L14
- https://github.com/mbta/skate/blob/14d9b756b4789ad28972d355fd9af4c8bd14dce3/lib/skate/settings/user_settings.ex#L31-L54

ex_cubic_ingestion/lib/ex_cubic_ingestion/downloader.ex
```diff
@@ -0,0 +1,144 @@
+defmodule ExCubicIngestion.Downloader do
```
grejdi-mbta, line 1: @paulswartz Was hoping to get some feedback on this module.

For some background and as discussed, I want to stream the download from Cubic (minimal memory footprint) and upload to our 'Incoming' bucket. This module represents the download piece and it's the more complicated part. You can see the rest in `FetchDmap.fetch_and_upload_to_s3`.

Although, download streaming worked pretty well with HTTPoison, the stream it was creating had very small chunks of binary in it. Unfortunately, the S3 upload stream did not accept these chunks as it has a minimum `5 * 1024 * 1024` byte size for chunks to be uploaded, see [this](https://github.com/ex-aws/ex_aws_s3/blob/v2.3.2/lib/ex_aws/s3/upload.ex#L71). This is a requirement by AWS, and so I needed to abide, hence the module below, where I essentially buffer chunks until I reach the minimum before sending it to the stream.

Please let me know what you think.

grejdi-mbta: I'd like to leave it in, as it gives us the ability to predict load more accurately, which can't always be said for data pipelines. If we bump concurrent worker to 5, we know that the system will be taking up ~25mb of memory. It also future proofs. 
Do you have any comments on the Elixir Stream aspect? Still very much new in this area, so want to make sure I didn't overlook anything important.

paulswartz: My main recommendation is to focus on getting the basics running correctly: we'll have time to come back and optimize this once the pipeline is working end-to-end. https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast

If you still want to continue down this path, some recommendations:
 - pull it into a new PR: I think the functionality you had before seemed close to being done
 - split up the "combine smaller packets into larger packets" from the "stream HTTP messages". `Stream` is the right interface for this functionality, and if they're separate, you can use `Stream.chunk_while` to handle the chunking into larger packets without needing to re-implement it alongside the HTTP streaming.
 - you'll want tests. basic tests could use https://httpbin.org/#/Dynamic_data/get_stream_bytes__n_ and https://httpbin.org/#/Dynamic_data/get_drip but there are a lot of edge cases: places when the connection could drop, packets can end exactly on a boundary, a packet could take longer than the timeout, &c

grejdi-mbta: Thanks @paulswartz. This is the feedback I was looking for. Very helpful.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_dmap_dataset.ex
grejdi-mbta: Applied. Thanks.

ex_cubic_ingestion/lib/ex_cubic_ingestion/downloader.ex
grejdi-mbta: Reverted to a simpler download/upload. Will create new PR with the stream solution.

grejdi-mbta: New PR here: https://github.com/mbta/data_platform/pull/51

grejdi-mbta:
> should the sample files have data?

Just wanted to have placeholders for now. I will add some in soon.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_dmap_feed.ex
```diff
@@ -0,0 +1,65 @@
...
+  def get!(id) do
+    Repo.get!(not_deleted(), id)
+  end
+
+  @doc """
+  Finds the dataset that was last updated and updates the feed's last updated value
+  """
+  @spec update_last_updated_from_datasets([CubicDmapDataset.t()], t()) :: t()
+  def update_last_updated_from_datasets(dataset_recs, rec) do
+    [latest_updated_dataset_rec | _rest] = Enum.sort_by(dataset_recs, & &1.last_updated_at, :desc)
```
paulswartz, line 57: If you only need the highest value, you can use `max_by`:
```suggestion
    latest_updated_dataset_rec = Enum.max_by(dataset_recs, & &1.last_updated_at, DateTime)
```

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
```diff
@@ -0,0 +1,26 @@
+defmodule ExCubicIngestion.Validators do
+  @moduledoc """
+  Module for holding helpful functions for validation
+  """
+
+  @spec is_valid_iso_date?(String.t()) :: boolean()
+  def is_valid_iso_date?(date_str) do
```
paulswartz, line 7:
```suggestion
  def valid_iso_date?(date_str) do
```

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
```diff
@@ -0,0 +1,26 @@
...
+    match?({:ok, _date}, Date.from_iso8601(date_str))
+  end
+
+  @spec is_valid_iso_datetime?(String.t()) :: boolean()
+  def is_valid_iso_datetime?(datetime_str) do
```
paulswartz, line 12:
```suggestion
  def valid_iso_datetime?(datetime_str) do
```

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
```diff
@@ -0,0 +1,26 @@
...
+    match?({:ok, _datetime}, Timex.parse(datetime_str, "{ISO:Extended}"))
+  end
+
+  @spec is_valid_dmap_dataset_url?(String.t()) :: boolean()
+  def is_valid_dmap_dataset_url?(url) do
```
paulswartz, line 17:
```suggestion
  def valid_dmap_dataset_url?(url) do
```

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
paulswartz: You can see the difference in the [Guard](https://hexdocs.pm/elixir/guards.html#why-guards) documentation:

```elixir
def empty_map?(map) when is_map(map), do: false
```

- the normal function has a `?` suffix but no `is_` prefix
- the guard has an `is_` prefix but not `?` suffix
```

File: ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex [hunk 14]
```diff
@@ -0,0 +1,26 @@
...
+    URI.parse(url).scheme == "https" &&
+      URI.parse(url).path not in [nil, "/"]
```
paulswartz, line 19:
```suggestion
    case URI.parse(url) do
        %{scheme: "https",
             path: valid_path} when is_binary(valid_path) and valid_path != "/" -> true
          %{} -> false
      end
```
alternatively, put the parsed `URI` into a variable instead of parsing it twice.

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/fetch_dmap.ex
```diff
@@ -0,0 +1,132 @@
...
+    |> CubicDmapFeed.update_last_updated_from_datasets(feed_rec)
+
+    :ok
+  end
+
+  @doc """
+  Make sure that the dataset has all the required fields and has valid data.
+  """
+  @spec is_valid_dataset(map()) :: boolean()
+  def is_valid_dataset(dataset) do
```
paulswartz, line 56:
```suggestion
  def valid_dataset?(dataset) do
```

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/fetch_dmap.ex
```diff
@@ -0,0 +1,132 @@
...
+  Construct the full URL to the feed, applying some overriding logic for
+  last updated (if passed in).
+  """
+  @spec construct_feed_url(CubicDmapFeed.t(), DateTime.t() | nil) :: String.t()
+  def construct_feed_url(feed_rec, last_updated \\ nil) do
+    dmap_base_url = Application.fetch_env!(:ex_cubic_ingestion, :dmap_base_url)
+
+    dmap_api_key = Application.fetch_env!(:ex_cubic_ingestion, :dmap_api_key)
+
+    last_updated_query_param =
```
paulswartz, line 81: Rather than duplicating the `Calendar` code, you could set the `last_updated` to the correct value if `feed_rec` has `last_updated_at`:

```elixir
last_updated = if feed_rec.last_updated_at and is_nil(last_updated) do
    DateTime.add(feed_rec.last_updated_at, 1, :microsecond)
else
    last_updated
end
lasted_updated_query_pararm = if last_updated do
  "&last_updated=#{Calendar.strftime(last_updated, "%Y-%m-%dT%H:%M:%S.%f")}"
else
  true
end
```

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/fetch_dmap.ex
paulswartz: question: should this live in `CubicDmapDataset`, since the actual parsing lives there?

ex_cubic_ingestion/test/ex_cubic_ingestion/workers/fetch_dmap_test.exs
```diff
@@ -0,0 +1,219 @@
...
+               FetchDmap.is_valid_dataset(dataset_empty)
+    end
+  end
+
+  describe "construct_feed_url/2" do
+    test "feed without a last updated timestamp" do
+      dmap_feed_relative_url = "/controlledresearchusersapi/sample"
+
+      dmap_feed =
+        Repo.insert!(%CubicDmapFeed{
```
paulswartz, line 123: question: do you need to insert the record into the database? I don't think the test case fetches anything else from the DB, or requires a primary key.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_dmap_feed.ex
grejdi-mbta: Done. Thanks. Good to know.

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
grejdi-mbta: Done.

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
grejdi-mbta: Opted for the variable.

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/fetch_dmap.ex
grejdi-mbta: There is a case where `feed_rec.last_updated_at` is `nil`, so tests were failing. I made a small adjustment to remove duplication. The code reads well to me, but I can see further optimization being made.

ex_cubic_ingestion/test/ex_cubic_ingestion/workers/fetch_dmap_test.exs
grejdi-mbta: Removed. I didn't like a schema object that doesn't actually represent a record, but it's all good.

paulswartz (approved):
🍰

PR #53, opened 2022-06-06
[elixir+db] feat: add EDW.SVN_INCIDENT ODS table rec for ingestion.

grejdi-mbta:
This PR adds the SVN_INCIDENT ODS table to be ingested. Table has already been added to the Glue Data Catalog with this deployment: https://github.com/mbta/devops/pull/487

### DIFF
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex
@@ -159,7 +159,7 @@ defmodule ExCubicIngestion.Workers.Ingest do
   @spec attach_ods_snapshot(map()) :: map()
   defp attach_ods_snapshot(%{partition_columns: partition_columns} = load) do
     if String.starts_with?(load[:s3_key], "cubic/ods_qlik/") do
-      ods_load_rec = CubicOdsLoadSnapshot.get_by!(load_id: load[:id])
+      ods_load_snapshot_rec = CubicOdsLoadSnapshot.get_by!(load_id: load[:id])
 
       # note: order of partitions is intentional
       %{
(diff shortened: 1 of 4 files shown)


ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex
```diff
@@ -159,15 +159,15 @@ defmodule ExCubicIngestion.Workers.Ingest do
   @spec attach_ods_snapshot(map()) :: map()
   defp attach_ods_snapshot(%{partition_columns: partition_columns} = load) do
     if String.starts_with?(load[:s3_key], "cubic/ods_qlik/") do
-      ods_load_rec = CubicOdsLoadSnapshot.get_by!(load_id: load[:id])
+      ods_load_snapshot_rec = CubicOdsLoadSnapshot.get_by!(load_id: load[:id])
```
grejdi-mbta, line 162: just a rename to make more accurate.

ex_cubic_ingestion/priv/repo/migrations/20220801144555_add_svn_incident_ods_table.exs
```diff
@@ -0,0 +1,25 @@
...
+      table_id: ods_edw_svn_incident_table_rec.id,
+      snapshot_s3_key: "cubic/ods_qlik/EDW.SVN_INCIDENT/LOAD00000001.csv.gz"
+    })
+  end
+
+  def down do
+    ods_edw_svn_incident_table_rec = CubicTable.get_by!(name: "cubic_ods_qlik__edw_svn_incident")
+    Repo.delete!(ods_edw_svn_incident_table_rec)
+    Repo.delete!(CubicOdsTableSnapshot.get_by!(table_id: ods_edw_svn_incident_table_rec.id))
+  end
```
paulswartz, line 24: question: do we also need to delete any relevant Load records?

grejdi-mbta: We could, but I rather not. I see the `down` really for reverting the `up` and nothing more.

grejdi-mbta: I understand. I just don't think it will happen and if it does, it will be something more critical. We will need to not just clean up loads from the DB, but also from S3 and other areas where they might end up. By deleting the loads with the down, we will not capture the full extent of rollback.

paulswartz (approved):
🍰

PR #54, opened 2022-06-09
[elixir] fix: further filter objects coming in for ingestion

grejdi-mbta:


paulswartz (commented):
What does the S3 data look like from the API? I'm surprised we didn't come across this issue before, and I thought the Delimiter part of the query caused prefixes to be returned separately?

grejdi-mbta:
> What does the S3 data look like from the API? I'm surprised we didn't come across this issue before, and I thought the Delimiter part of the query caused prefixes to be returned separately?

They look like regular objects, but with a size of 0, and the key being the folder path. We didn't run into it because we mostly moved objects in through `aws s3 cp` or some other mean where the actual files were being moved/copied. What Cubic/Qlik did is that they created the folder first and then uploaded. It's similar to using the 'Create Folder' button on the S3 AWS console.

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
```diff
@@ -24,4 +24,10 @@ defmodule ExCubicIngestion.Validators do
   def map_has_keys?(map, key_list) do
     Enum.all?(key_list, &Map.has_key?(map, &1))
   end
+
+  @spec valid_s3_object?(map()) :: boolean()
+  def valid_s3_object?(object) do
+    Map.has_key?(object, :key) && not String.ends_with?(object.key, "/") &&
+      Map.has_key?(object, :size) && object.size != "0"
+  end
```
paulswartz, line 32: suggestion: this might be better as a pattern match
```suggestion
  def valid_s3_object?(object)
  def valid_s3_object?(%{key: key, size: size}) when size != "0" do
    not String.ends_with?(key, "/")
  end
  def valid_s3_object?(%{}) do
    false
  end
```

ex_cubic_ingestion/test/ex_cubic_ingestion/validators_test.exs
```diff
@@ -0,0 +1,113 @@
...
+    test "invalid if missing size" do
+      object = %{
+        # ...
+        key: "cubic/ods_qlik/EDW.SAMPLE/LOAD1.csv"
+      }
+
+      refute Validators.valid_s3_object?(object)
+    end
+
+    test "invalid if size is 0" do
```
paulswartz, line 103: question: I thought the only invalid files were those which ended in "/" *and* had a 0 size?

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
grejdi-mbta: Fixed.

ex_cubic_ingestion/test/ex_cubic_ingestion/validators_test.exs
grejdi-mbta: Done. Will be now sending them to 'ready_for_erroring' so they can be moved out of incoming.

paulswartz (approved):
🍰

PR #55, opened 2022-06-11
[doc] feat: ADR for standardizing all Cubic data to string type

grejdi-mbta:
This PR adds an ADR to makes the case for standardizing the Incoming bucket schema columns to all be `string` initially. We're hoping to stabilize the ingestion errors, and only start makings adjustment to the Springboard schemas after we understand the data better.

### DIFF
--- a/doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
+++ b/doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
@@ -1,5 +1,5 @@
 
-# 7. Standardizing on 'string' for all Glue Catalog Table column types
+# 7. Standardizing on 'string' for all initial Glue Catalog Table column types
 
 Date: 2022-06-09
 
@@ -30,7 +30,7 @@ s3://mbta-ctd-dataplatform-dev-springboard/cubic/ods_qlik/EDW.ABP_TAP
 to org.apache.parquet.io.PrimitiveColumnIO
 ```
 
-Some stakeholders will be pulling data right out of the Springboard bucket. We need to make sure that it's always available. We obviously can't guarantee this, but we can try by defaulting on the `string` type for all columns, and only swaying from the default when we are sure that we can determine the type and have the proper fallbacks in place. Ideally, a `string` type for a column of data in Incoming bucket with `1`, `2`, `3` as values can be an `int` type in the Springboard. These conversions do make Athena querying more efficient, so we should strive to apply them as much as possible, but for unknown data we should default to using the `string` type.
+Some stakeholders will be pulling data right out of the Springboard bucket. We need to make sure that it's always available. We obviously can't guarantee this, but we can try by defaulting on the `string` type for all columns, and only swaying from the default when we are sure that we can determine the type and have the proper fallbacks in place. Ideally, a `string` type for a column of data in Incoming bucket with `1`, `2`, `3` as values can be an `bigint` type in the Springboard. These conversions do make Athena querying more efficient, so we should strive to apply them as much as possible, but for unknown data we should default to using the `string` type.
 
 ## Assumptions
 
@@ -38,25 +38,34 @@ AWS Glue is still a fairly new technology and it seems to be constantly evolving
 
 Unfortunately Athena and Glue are not open source and therefore we cannot make an absolute assertion that this is occuring. We are only relying on experience with this new technology (in Glue's case) to make this assumption.
 
-In addition, we are also assuming that standardizing on the type, specifically the `string` type, is a good approach to dealing with data querying issues in Athena. Almost all types have a `toString()` implementation, so relying on it feels safe. It also seems to be a fallback for the Glue Crawler, as we tend to only see `string`  and `'bigint'` types surface from its process.
+In addition, we are also assuming that initially standardizing on the type, specifically the `string` type, is a good approach to dealing with data querying issues in Athena. Almost all types have a `toString()` implementation, so relying on it feels safe. It also seems to be a fallback for the Glue Crawler, as we tend to only see `string`  and `'bigint'` types surface from its process.
 
 ## Implemetation
 
-As indicated, our solution will involve standardizing the type for all data coming in. We will choose the `string` type, as it will be the most inclusive and is usually supported in a standard way across many technologies. This will minimize any issues with being able to write the data to parquet and reading it out.
+As indicated, our solution will involve initially standardizing the type for all data coming in. We will choose the `string` type, as it will be the most inclusive and is usually supported in a standard way across many technologies. This will minimize any issues with being able to write the data to parquet and reading it out.
 
 Our implementation will be as follows:
 
-* For all Glue Catalog Tables currently defined, we will update their types to be `string`. We then have 2 options to reprocess the existing data:
+1. Update Glue table Terraform modules to identify all incoming data columns as `string` types.
+
+2. Add an additional table to the Terraform modules for the Springboard destination, which will be marked with the 'raw_' prefix (ex. `raw_cubic_ods_qlik__edw_sample`), and will have a location that includes the 'raw/' prefix (ex. `s3://springboard-bucket/raw/cubic/ods_qlik/EDW.SAMPLE/...`). This table will be used to query un-processed data with Athena. (Note: The existing Springboard destination table can be used only once some types have been determined and the data is being ingested and processed.)
+
+3. In order to keep track that the loads being ingested are processed or not, we will mark the table and load records with the `is_raw` flag. This will reflect the state of the load in the Springboard bucket.
+
+4. Make adjustments to the Glue Job to allow 'raw' loads to be stored with a different prefix in the Springboard bucket.
+
+5. Reprocess the existing data:
 
     - Ask Cubic to restart Qlik and send new snapshots for all ODS tables to the Incoming bucket and on to the Springboard. We can then clean up the Springboard bucket and/or delete older partitions from Athena.
     - Reprocess the data ourselves utilizing the Archive bucket and overwrite the Springboard data. We will then reload partitions in Athena, which might not be necessary.
 
-* For all new Glue Catalog Tables, we will set the column types to `string`.
+## Moving Forward
 
-Note: Once this practice is established, we can then start analyzing the data and making an accurate determination of the types. These determinations will be done with fallbacks in mind, as in nullifying fields if data is not of the correct type, or applying transformations and/or machine learning techniques for completing values.
+1. All new tables added for ingestion will be marked with `is_raw: true` initially.
+
+2. Once some or all column types have been determined for the table, we will switch the `is_raw` flag to `false`.
 
 ## Consequences
 
-* Even though standardizing on the `string` type will create the stability we are seeking, we have to be conscious of the fact that we are not really moving the needle on cleaning up the data and truly preparing it for further ingestion. Currently, we don't have visibility on how this data will be ingested beyond being queried in Athena and possibly downloaded there, so it feels premature to make optimization beyond this standard. However, with this architecture and base we can easily expand scope and achieve results as noted above. We have yet to tap into the many more features Glue has advertised.
-* We have created a module for Glue tables (see [this](https://github.com/mbta/devops/tree/master/terraform/modules/app-dataplatform/glue-data-catalog/cubic_ods_qlik__table) and [this](https://github.com/mbta/devops/tree/master/terraform/modules/app-dataplatform/glue-data-catalog/cubic_dmap__table)). If we execute on this ADR, we will need to separate the schemas for Incoming and Springboard, since some columns in Incoming will be `string` and will be of other types in Springboard.
-
+* Even though initially standardizing on the `string` type will create the stability we are seeking, we have to be conscious of the fact that we are not really moving the needle on cleaning up the data and truly preparing it for further ingestion. Currently, we don't have visibility on how this data will be ingested beyond being queried in Athena and possibly downloaded there, so it feels premature to make optimization beyond this standard. However, with this architecture and base we can easily expand scope and achieve results as noted above.
+* We will be creating additional tables in the Glue Databases in the Terraform modules (see [this](https://github.com/mbta/devops/tree/master/terraform/modules/app-dataplatform/glue-data-catalog/cubic_ods_qlik__table) and [this](https://github.com/mbta/devops/tree/master/terraform/modules/app-dataplatform/glue-data-catalog/cubic_dmap__table)). Although cost-wise this will not have much impact, we might have tables in Glue that are not being utilized, hence creating some clutter.
--- a/doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
+++ b/doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
@@ -30,31 +30,33 @@ s3://mbta-ctd-dataplatform-dev-springboard/cubic/ods_qlik/EDW.ABP_TAP
 to org.apache.parquet.io.PrimitiveColumnIO
 ```
 
-Some stakeholders will be pulling data right out of the Springboard bucket. We need to make sure that it's always available. We obviously can't guarantee this, but we can try by standardizing on the `'string'` type for all columns, and only swaying from the standard when we are absolutely sure and have the proper fallbacks in place.
+Some stakeholders will be pulling data right out of the Springboard bucket. We need to make sure that it's always available. We obviously can't guarantee this, but we can try by defaulting on the `string` type for all columns, and only swaying from the default when we are sure that we can determine the type and have the proper fallbacks in place. Ideally, a `string` type for a column of data in Incoming bucket with `1`, `2`, `3` as values can be an `int` type in the Springboard. These conversions do make Athena querying more efficient, so we should strive to apply them as much as possible, but for unknown data we should default to using the `string` type.
 
 ## Assumptions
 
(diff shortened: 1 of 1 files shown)


doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,60 @@
...
+
+In addition, we are also assuming that standardizing on the type, specifically the `'string'` type, is a good approach to dealing with data querying issues in Athena. Almost all types have a `toString()` implementation, so relying on it feels safe. It also seems to be a fallback for the Glue Crawler, as we tend to only see `'string'`  and `'bigint'` types surface from its process.
+
+## Implemetation
+
+As indicated, our solution will involve standardizing the type for all data coming in. We will choose the `'string'` type, as it will be the most inclusive and is usually supported in a standard way across many technologies. This will minimize any issues with being able to write the data to parquet and reading it out.
+
+Our implementation will be as follows:
+
+* For all Glue Catalog Tables currently defined, we will update their types to be `'string'`. For reprocessing the data, we have 2 options:
```
grejdi-mbta, line 49: This implementation has been tested and worked for the issues indicated above.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,60 @@
...
+
+Date: 2022-06-09
+
+## Status
+
+Draft
+
+## Context
+
+We have been using the AWS Glue Crawler to determine the columns and types of data that are coming from Cubic for both ODS and DMAP data feeds. In most cases, the Crawler will sample the data and make the determination. This has worked pretty well in quickly getting the column names, but not that great for figuring out the types of these columns. For some tables with a small amount of data, such as `cubic_ods_qlik__edw_benefit_status_dimension`, the sample size is consitent and many times will not change, resulting in the crawler making an accurate assumption about the types. But for much larger tables, the sample size may be randomly determined causing the crawler to make mistake as to what the types of the columns are (see Assumptions).
```
paulswartz, line 12: question: I didn't think we were using the Glue Crawler? I thought we were providing an explicit schema for each `incoming` and `springboard` table via Terraform?

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,60 @@
...
+```
+HIVE_CANNOT_OPEN_SPLIT: Error opening Hive split 
+s3://mbta-ctd-dataplatform-dev-springboard/cubic/ods_qlik/EDW.ABP_TAP
+/snapshot=20220607T160603Z/identifier=LOAD00000001.csv.gz
+/part-00015-7cc52de2-9ca3-4ede-b2b7-004cea96a414.c000.snappy.parquet 
+(offset=0, length=2904265): org.apache.parquet.io.GroupColumnIO cannot be cast 
+to org.apache.parquet.io.PrimitiveColumnIO
+```
+
+Some stakeholders will be pulling data right out of the Springboard bucket. We need to make sure that it's always available. We obviously can't guarantee this, but we can try by standardizing on the `'string'` type for all columns, and only swaying from the standard when we are absolutely sure and have the proper fallbacks in place.
```
paulswartz, line 33: suggestion: I think standardizing on `string` for the `incoming` bucket is fine, but I think we should be writing Parquet files with the correct schemas. It's more efficient to write and to query, and we'll get notified earlier if there's an error with a file not matching our expected schema.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Assumptions
+
+AWS Glue is still a fairly new technology and it seems to be constantly evolving. We believe that the Crawler is possibly making mistakes. For a non-specific example, let's imagine a column containing `'1'` for the first row, but `''` for the next 100 rows. If the crawler looks at the first 10 rows, it might make the determination that it is a `'bigint'` type column, but if it looks at the last 10 rows, it might make the determination that it's a `'string'` type column.
```
paulswartz, line 37: - I'd agree that it's constantly evolving, but it's been available for almost 5 years.
- [How Glue does schema inference](https://aws.amazon.com/premiumsupport/knowledge-center/glue-crawler-detect-schema/) is described in their Knowledge Center:
> During the first crawler run, the crawler reads either the first 1,000 records or the first megabyte of each file to infer the schema. The amount of data read depends on the file format and availability of a valid record. [...] For CSV files, the crawler reads either the first 1000 records or the first 1 MB of data, whatever comes first.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,60 @@
...
+
+Unfortunately Athena and Glue are not open source and therefore we cannot make an absolute assertion that this is occuring. We are only relying on experience with this new technology (in Glue's case) to make this assumption.
+
+In addition, we are also assuming that standardizing on the type, specifically the `'string'` type, is a good approach to dealing with data querying issues in Athena. Almost all types have a `toString()` implementation, so relying on it feels safe. It also seems to be a fallback for the Glue Crawler, as we tend to only see `'string'`  and `'bigint'` types surface from its process.
```
paulswartz, line 41: I've also seen problems with the crawler's finding of `bigint` columns. It will detect IDs from Cubic as `bigint` even if the values are larger than what a signed 64-bit integer can store.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
paulswartz, line 49: I agree with this approach for the `incoming` tables: at the end of the data CSV values are all strings.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Implemetation
+
+As indicated, our solution will involve standardizing the type for all data coming in. We will choose the `'string'` type, as it will be the most inclusive and is usually supported in a standard way across many technologies. This will minimize any issues with being able to write the data to parquet and reading it out.
+
+Our implementation will be as follows:
+
+* For all Glue Catalog Tables currently defined, we will update their types to be `'string'`. For reprocessing the data, we have 2 options:
+    - Ask Cubic to restart Qlik and sending a new batch of files to the Incoming bucket and on to the Springboard. We can then clean up the Springboard bucket and/or delete faulty partitions from Athena.
```
paulswartz, line 50: question: are we able to do this ourselves by using the Archive?

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
grejdi-mbta: We are using it to get the schemas quickly, and then transferring them to TF for future management and documentation.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
grejdi-mbta: Yes. The second bullet is the other option and is what you are describing. I'll make it more clear.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
grejdi-mbta: Updated to reference this article and added to the assumption.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
grejdi-mbta: Cleared it up a little. Hope it helps.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
grejdi-mbta: Also, added a consequence.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,62 @@
...
+```
+HIVE_CANNOT_OPEN_SPLIT: Error opening Hive split 
+s3://mbta-ctd-dataplatform-dev-springboard/cubic/ods_qlik/EDW.ABP_TAP
+/snapshot=20220607T160603Z/identifier=LOAD00000001.csv.gz
+/part-00015-7cc52de2-9ca3-4ede-b2b7-004cea96a414.c000.snappy.parquet 
+(offset=0, length=2904265): org.apache.parquet.io.GroupColumnIO cannot be cast 
+to org.apache.parquet.io.PrimitiveColumnIO
+```
+
+Some stakeholders will be pulling data right out of the Springboard bucket. We need to make sure that it's always available. We obviously can't guarantee this, but we can try by defaulting on the `string` type for all columns, and only swaying from the default when we are sure that we can determine the type and have the proper fallbacks in place. Ideally, a `string` type for a column of data in Incoming bucket with `1`, `2`, `3` as values can be an `int` type in the Springboard. These conversions do make Athena querying more efficient, so we should strive to apply them as much as possible, but for unknown data we should default to using the `string` type.
```
paulswartz, line 33: Do we have "unknown" data in the Springboard? There's a document with the types of all the columns, so we should know the expected type of everything, even if it's coming in as a "string" in the CSV.

grejdi-mbta: I do have the document. It seems to be auto-generated HTML. The columns on there have types and indication of length, nullability, etc., but frankly I'm not sure how up-to-date it is and whether it makes sense to follow it. One of the reasons I can think of is that these types are defined in another system, not in Glue or Athena. We'd have to analyze and map them correctly to Glue and Athena. The other reason is my hesitancy that Cubic will keep it up-to-date even if we ask. As it stands, a few of the tables listed there have never made it through. Others have made it through but [have no data](https://app.asana.com/0/546643571451005/1202246207611231). And for some of those that have made it in, the columns are not correct. So when you say "unknown", the short answer is Yes.

grejdi-mbta:
> Continuing to think about this: are there issues with the schemas other than some `bigint` columns being incorrectly determined?
> 
> If not, the fix seems to be smaller: correctly labeling non-`bigint` columns in our schema definitions, but without making more general changes to the schema.

Here is a rough breakdown of the ~3790 columns currently:
```
32 double
1239 bigint
2523 string
```
The `bigint` issue is pretty much __the__ issue. I believe a lot of them are mislabeled, but I cannot be sure without painstakingly going through them. 

With this ADR, I'd like to create some stability in this system. Without making a decision like this, we'll be chasing these issues for a long time especially since Cubic has full control of the data that they send. I plan on asking how we can establish a process for changes in our weekly meeting, but I don't expect to have a clear answer.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,62 @@
...
+
+Date: 2022-06-09
+
+## Status
+
+Draft
+
+## Context
+
+We have been using the AWS Glue Crawler to determine the columns and types of data that are coming from Cubic for both ODS and DMAP data feeds. In most cases, the Crawler will sample the data and make the determination. This has worked pretty well in quickly getting the column names, but not that great for figuring out the types of these columns. For some tables with a small amount of data, such as `cubic_ods_qlik__edw_benefit_status_dimension`, the sample size is consitent and many times will not change, resulting in the crawler making an accurate assumption about the types. But for much larger tables, the sample size may be randomly determined causing the crawler to make mistake as to what the types of the columns are (see Assumptions).
```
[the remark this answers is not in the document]
grejdi-mbta: Although I think you are right that the `bigint` issue is separate, I do think that for smaller tables the crawler seems to work better. No direct evidence though, just that with some small tables I don't have an issue with querying them in Athena.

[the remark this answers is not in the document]
grejdi-mbta: Updated.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
paulswartz: > I do have the document. It seems to be auto-generated HTML.
It's auto-generated from the ODS database schema, as far as I can tell. If there's a better format to receive the schema, we should ask for it. (maybe a DDL, or a structured file?).

> One of the reasons I can think of is that these types are defined in another system, not in Glue or Athena. 
I don't think there's a way around the fact that Cubic defines the schema of the incoming data. My goal for the processing from Incoming -> Springboard is that the Springboard has well-typed schema, regardless of what Cubic is providing. 

If all we are doing is giving Springboard users access to the stringly-typed CSV data, then we don't need to do any processing: we could copy the files from Incoming -> Springboard and be done.

> The other reason is my hesitancy that Cubic will keep it up-to-date even if we ask.
If we're concerned that Cubic won't keep a schema up-to-date, that sounds like even more reason to not treat everything as a string. If a column changes from an integer to something else, I don't think we'd want to continue processing that table without some additional changes on our end: we'd want those imports to fail.

> And for some of those that have made it in, the columns are not correct.
Is there a ticket for these? If their information about the tables is incorrect, that is an issue Cubic should be addressing, either by updating the documentation or by updating the tables.

grejdi-mbta: > It's auto-generated from the ODS database schema, as far as I can tell. If there's a better format to receive the schema, we should ask for it. (maybe a DDL, or a structured file?).

Added [Asana ticket](https://app.asana.com/0/546643571451005/1202452782579116) for Cubic to look into.

> I don't think there's a way around the fact that Cubic defines the schema of the incoming data. My goal for the processing from Incoming -> Springboard is that the Springboard has well-typed schema, regardless of what Cubic is providing.

Agreed, and I'd like to get there. We will have to go through each table and analyze them to determine correct types in Athena/Glue. I can go down this path and it will be tedious, but this effort might not be worth for all tables and it will be a on-going continuous effort. I rather we make this effort towards the tail-end rather than upfront because it will matter for some and it won't for others, and we will only know what those are when the data gets consumed.

> If all we are doing is giving Springboard users access to the stringly-typed CSV data, then we don't need to do any processing: we could copy the files from Incoming -> Springboard and be done.

This is already provided by the Archive. We can't simply copy because of the snapshot, but yes, you're right. The goal of involving Glue is to also gain experience with it. See what it can handle, and ultimately the transformations it can do.

> If we're concerned that Cubic won't keep a schema up-to-date, that sounds like even more reason to not treat everything as a string. If a column changes from an integer to something else, I don't think we'd want to continue processing that table without some additional changes on our end: we'd want those imports to fail.

I am concerned with that, and I am not confident that they will stick to whatever process we establish. But sending everything into the Error bucket is not the answer. We would be erroring for technical reasons instead of requirements reasons. If we have enforced a type on a field that no ones cares about and our pipeline is failing because of that, then we have failed our end user who is relying on the data to come through.

> Is there a ticket for these? If their information about the tables is incorrect, that is an issue Cubic should be addressing, either by updating the documentation or by updating the tables.

Yes. I'm gathering more as I go. Here is [one](https://app.asana.com/0/546643571451005/1202246207611231) with missing tables and the other one is an agenda item. Once I have more information on how Cubic wants them, I will have more.

paulswartz: > We would be erroring for technical reasons instead of requirements reasons. If we have enforced a type on a field that no ones cares about and our pipeline is failing because of that, then we have failed our end user who is relying on the data to come through.

Maybe this is a question for OPMI then, as our main stakeholder? Would they prefer to handle possibly mis-typed data on their end, but always have the latest data? Or have the data processing stop while the types are corrected?

grejdi-mbta: This is an on-going thread with Product, so closing it out.

grejdi-mbta:
@paulswartz I've updated the ADR to reflect the discussion we had on Thursday. Here is some supporting PRs that I made as proof of concepts:

* Casting that throw exceptions, which support non-raw processing of tables. See: https://github.com/mbta/data_platform/pull/64
* Adding 'raw' indication for table that will not be processed: https://github.com/mbta/data_platform/pull/64 and https://github.com/mbta/devops/pull/453

I will refine these PRs once we agree on this ADR.

doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md
```diff
@@ -0,0 +1,71 @@
...
+5. Reprocess the existing data:
+
+    - Ask Cubic to restart Qlik and send new snapshots for all ODS tables to the Incoming bucket and on to the Springboard. We can then clean up the Springboard bucket and/or delete older partitions from Athena.
+    - Reprocess the data ourselves utilizing the Archive bucket and overwrite the Springboard data. We will then reload partitions in Athena, which might not be necessary.
+
+## Moving Forward
+
+1. All new tables added for ingestion will be marked with `is_raw: true` initially.
+
+2. Once some or all column types have been determined for the table, we will switch the `is_raw` flag to `false`.
```
paulswartz, line 66: question: Will we switch the flag, or create a new `CubicTable` with `is_raw` false? Otherwise, won't this prevent any new data from getting into the raw-prefixed table?

grejdi-mbta: Once we have prepped the types for the columns, we will switch the flag on the table. After the switch, all future loads will be ingested with the non-raw schema. We can then re-process the loads prior to the switch. Or have cubic restart Qlik.

paulswartz (approved):

grejdi-mbta:
@paulswartz Thanks. I'll be addressing issues in the linked PRs, so look out for review requests for those.

PR #56, opened 2022-06-11
[elixir] feat: ability to execute a schedule DMAP fetching upon running the task.

grejdi-mbta:
This PR serves as a proof of concept for running Cubic Ingestion tasks. I'm hoping to use what I learn here for attempting retries of glue jobs that have failed.

### DIFF
--- a/ex_cubic_ingestion/config/config.exs
+++ b/ex_cubic_ingestion/config/config.exs
@@ -8,7 +8,7 @@ config :logger, :console,
 config :ex_cubic_ingestion,
   ecto_repos: [ExCubicIngestion.Repo],
   run_migrations_at_startup?: false,
-  start_app?: true
+  start_app_children?: true
 
 config :ex_cubic_ingestion, Oban,
   repo: ExCubicIngestion.Repo,
--- a/ex_cubic_ingestion/lib/mix/tasks/schedule_dmap.ex
+++ b/ex_cubic_ingestion/lib/mix/tasks/schedule_dmap.ex
@@ -1,0 +1,25 @@
+defmodule Mix.Tasks.ScheduleDmap do
+  @moduledoc """
+  This task will insert the ScheduleDmap job upon execution, resulting in
+  the fetching of all DMAP feeds.
+  """
+  use Mix.Task
+
+  alias ExCubicIngestion.Workers.ScheduleDmap
+
+  require Oban
+
+  @doc """
+  Start the application without the GenServers, so we don't have competing
+  processes for querying the incoming bucket, and then inserting a ScheduleDMAP
+  job for Oban to process.
+  """
+  @impl Mix.Task
+  def run(_args) do
+    Application.put_env(:ex_cubic_ingestion, :start_app_children?, false)
+
+    Mix.Task.run("app.start")
+
+    Oban.insert!(ScheduleDmap.new(%{}))
+  end
+end
--- a/ex_cubic_ingestion/config/test.exs
+++ b/ex_cubic_ingestion/config/test.exs
@@ -4,7 +4,7 @@ import Config
 config :logger, level: :warning
 
 config :ex_cubic_ingestion,
-  start_app?: false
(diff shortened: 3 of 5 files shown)


ex_cubic_ingestion/config/config.exs
```diff
@@ -8,7 +8,7 @@ config :logger, :console,
 config :ex_cubic_ingestion,
   ecto_repos: [ExCubicIngestion.Repo],
   run_migrations_at_startup?: false,
-  start_app?: true
+  start_app_children?: true
```
grejdi-mbta, line 11: Gave this a more specific name.

grejdi-mbta:
@paulswartz let me know what you think about this approach for running tasks. I'm hoping to use this in AWS Batch or ECS Run Standalone Tasks with [this](https://github.com/mbta/actions/tree/main/run-ecs-fargate-task).

grejdi-mbta:
@paulswartz Ping, in case it slipped through.

ex_cubic_ingestion/lib/mix/tasks/schedule_dmap.ex
```diff
@@ -0,0 +1,25 @@
...
+  processes for querying the incoming bucket, and then inserting a ScheduleDMAP
+  job for Oban to process.
+  """
+  @impl Mix.Task
+  def run(_args) do
+    Application.put_env(:ex_cubic_ingestion, :start_app_children?, false)
+
+    Mix.Task.run("app.start")
+
+    Oban.insert(ScheduleDmap.new(%{}))
```
paulswartz, line 23: question: what happens if the insert fails?

paulswartz (commented):
thanks, I did miss this!

ex_cubic_ingestion/lib/mix/tasks/schedule_dmap.ex
grejdi-mbta: An exception will happen and the job will not get inserted. I'm assuming depending on where it's going to be run (AWS Batch or ECS) the log should indicate that it failed to run.

grejdi-mbta:
@paulswartz also let me know about this comment: https://github.com/mbta/data_platform/pull/56#issuecomment-1152980566

ex_cubic_ingestion/lib/mix/tasks/schedule_dmap.ex
paulswartz: I don't think that's the case: https://hexdocs.pm/oban/Oban.html#insert/2 says that it returns a tuple rather than raising an exception. Did you mean to use [`insert!`](https://hexdocs.pm/oban/Oban.html#insert!/2)?
```suggestion
    Oban.insert!(ScheduleDmap.new(%{}))
```

paulswartz:
> @paulswartz let me know what you think about this approach for running tasks. I'm hoping to use this in AWS Batch or ECS Run Standalone Tasks with [this](https://github.com/mbta/actions/tree/main/run-ecs-fargate-task).

Are you thinking that this will be used for one-off runs of the DMAP ingestion?

grejdi-mbta:
> > @paulswartz let me know what you think about this approach for running tasks. I'm hoping to use this in AWS Batch or ECS Run Standalone Tasks with [this](https://github.com/mbta/actions/tree/main/run-ecs-fargate-task).
> 
> Are you thinking that this will be used for one-off runs of the DMAP ingestion?

Yes. The ScheduleDmap job will be run through Oban Cron, see [this](https://github.com/mbta/data_platform/blob/main/ex_cubic_ingestion/config/config.exs#L19). This is a side task, in case we need to run this at any other time.

ex_cubic_ingestion/lib/mix/tasks/schedule_dmap.ex
grejdi-mbta: Thanks. Updated.

paulswartz (approved):
🍰

PR #57, opened 2022-06-13
[elixir] bug: allow empty datasets to be passed into dmap feed update

grejdi-mbta:
When there is no update to the feed, the worker will pass an empty list of datasets to the update function of DMAP feed. This PR handles this case, so `max_by/3` doesn't error out.


paulswartz (approved):
🍰

PR #64, opened 2022-06-21
[python+elixir] type casting for springboard

grejdi-mbta:
This PR adds logic to cast columns to supported types in the Springboard bucket. It introduces custom UDFs for Spark, so we can throw exceptions if there is an issue with type casting.


py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -8,7 +8,22 @@
...
 from pyspark.sql.dataframe import DataFrame
-from pyspark.sql.functions import lit
+from pyspark.sql.functions import lit, col
+import boto3
+
+# helper variables
+glue_client = boto3.client('glue')
+
+# Ref: https://github.com/MrPowers/spark-daria/blob/main/src/main/scala/com/github/mrpowers/spark/daria/sql/DataFrameHelpers.scala#L198
+athena_type_to_spark_type = {
```
grejdi-mbta, line 18: This is not the best map and will continue to work on it. Some of it might be off, such as the 'decimal' type.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -109,6 +161,32 @@ def from_catalog_kwargs(load: dict, env: dict) -> dict:
...
+
+    Returns
+    -------
+    DataFrame
+        Updated DataFrame containing an updated schema
+    """
+
+    column_statements = []
+    for field in schema_fields:
+        column_statements.append(f'cast ({field["name"]} as {field["type"]}) as {field["name"]}')
```
grejdi-mbta, line 185: Where the casting happens. Note that the field will set to NULL if unable to cast.

py_cubic_ingestion/pyproject.toml
```diff
@@ -18,6 +18,7 @@ mypy = "^0.941"
 coverage = "6.3.2"
 pylint = "^2.13.3"
 pyspark = "3.1.1"
+boto3 = "^1.24.13"
```
grejdi-mbta, line 21: Added boto3 for Glue client.

py_cubic_ingestion/tests/test_job_helpers.py
```diff
@@ -195,6 +272,91 @@ def test_from_catalog_kwargs() -> None:
...
+    """
+    Test creating a DataFrame with an updated schema and that
+    the data is valid.
+
+    Parameters
+    ----------
+    spark_session : list
+        Fixture that contains the Spark Session to use
+    """
+    original_data = [
```
grejdi-mbta, line 285: In the future, I'd like to expand this list as a way to validate data casting.

py_cubic_ingestion/lib/py_cubic_ingestion/custom_udfs.py
```diff
@@ -0,0 +1,55 @@
...
+            return None
+
+        # int is the type for long as well
+        val = int(s)
+        if val < -9223372036854775808 or val > 9223372036854775807:
+            raise ValueError(f"'{s}', Out of range of Spark LongType data type. Use DecimalType.")
+
+        return val
+    except ValueError as e:
+        raise ValueError(f"'{s}', {e}") from e
```
grejdi-mbta, line 27: Re-raising so we can add the value to the exception.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -3,15 +3,35 @@
...
 import logging
 import json
-import typing
-from pyspark.sql.dataframe import DataFrame
-from pyspark.sql.functions import lit
+import os
+
+
+# helper variables
+glue_client = boto3.client("glue", region_name="us-east-1")
```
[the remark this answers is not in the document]
paulswartz: I figured boto3 would pull from `AWS_REGION`?

docker/Dockerfile.glue_2_0__local
```diff
@@ -36,7 +36,7 @@ RUN pip install poetry
 COPY ./py_cubic_ingestion /data_platform/py_cubic_ingestion
 WORKDIR /data_platform/py_cubic_ingestion
 RUN poetry config virtualenvs.create false
-RUN poetry install --no-dev
+RUN poetry install
```
paulswartz, line 39: question: does this mean a dev dependency should become a real dependency?

py_cubic_ingestion/lib/py_cubic_ingestion/custom_udfs.py
```diff
@@ -0,0 +1,55 @@
...
+Custome Spark user-defined functions (UDFs) for processing dataframes.
+"""
+
+from datetime import date, datetime
+from typing import Optional
+from dateutil import parser
+from dateutil.tz import UTC
+
+
+def is_empty(s: Optional[str]) -> bool:
```
paulswartz, line 11: question: the uses here all seem like they're providing a `str` and not None?

py_cubic_ingestion/lib/py_cubic_ingestion/custom_udfs.py
```diff
@@ -0,0 +1,55 @@
...
+    return not (s and not s.isspace())
+
+
```
paulswartz, line 14: suggestion: I don't think all of our columns are possibly empty, so including `""` as a possible value doesn't seem quite correct. it also duplicates the optional parsing. what about pulling the optional handling into a decorator?

```python
T = TypeVar("T")
def as_optional(parser: Function[[str], T) -> Function[[str], Optional[T]:
  def internal(s: str) -> Optional[T]:
    if is_empty(s):
      return None
    return parser(s)
  return internal

def as_long(s: str) -> int:
  ...

as_optional_long: Function[[str], Optional[int]] = as_optional(as_long)
as_optional_date: function[[str], Optional[datetime]] = as_optional(as_date)
```

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -109,6 +165,43 @@ def from_catalog_kwargs(load: dict, env: dict) -> dict:
...
+        if field["type"] == "long":
+            column_statements.append(as_long_udf(field["name"]).alias(field["name"]))
+        elif field["type"] == "double":
+            column_statements.append(as_double_udf(field["name"]).alias(field["name"]))
+        elif field["type"] == "date":
+            column_statements.append(as_date_udf(field["name"]).alias(field["name"]))
+        elif field["type"] == "timestamp":
+            column_statements.append(as_timestamp_udf(field["name"]).alias(field["name"]))
+        else:
+            column_statements.append(field["name"])
```
paulswartz, line 198: suggestion: this feels like there should be a dict mapping `type` to the UDF.

py_cubic_ingestion/pyproject.toml
paulswartz: question: (per above) shouldn't this be a real dependency?

py_cubic_ingestion/tests/test_job_helpers.py
```diff
@@ -195,6 +243,190 @@ def test_from_catalog_kwargs() -> None:
...
+    updated_df = job_helpers.df_with_updated_schema(
+        original_df,
+        [
+            {"name": "double_col", "type": "double"},
+            {"name": "double_col_error", "type": "double"},
+        ],
+    )
+
+    with pytest.raises(PythonException) as excinfo:
+        updated_df.first()
```
paulswartz, line 367: question: why does the exception happen when getting the record, rather than when updating the DataFrame?

docker/Dockerfile.glue_2_0__local
grejdi-mbta: In the Glue environment, boto3 is already included. See: https://docs.aws.amazon.com/glue/latest/dg/aws-glue-programming-python-libraries.html

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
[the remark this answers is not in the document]
grejdi-mbta: Wasn't aware that this might exist with Glue. I knew it exists in Lambdas. I'll double check and add.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
grejdi-mbta: I'm not a big fan of using dicts this way. I prefer their usage only in data structures.

py_cubic_ingestion/pyproject.toml
grejdi-mbta: As noted, it's already included in Glue. I guess I can add it. It just makes the package that Glue needs to download a  little bigger.

py_cubic_ingestion/tests/test_job_helpers.py
grejdi-mbta: Spark is 'lazy' and only executes operations such as casting when actually working with the record, as in getting the record or writing it out to another format.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
paulswartz: You're already doing it earlier in the file:
https://github.com/mbta/data_platform/pull/64/files#diff-a322a77420acbc9a7fa72b72c7d6836290ad9dc8ecb7ee7dabd74b000d0ddb47R20-R26
https://github.com/mbta/data_platform/pull/64/files#diff-a322a77420acbc9a7fa72b72c7d6836290ad9dc8ecb7ee7dabd74b000d0ddb47R129

Otherwise, could you pull the duplicated parts out?
```python
for field in schema_fields:
  udf = None
  if field["type"] == "long":
    udf = as_long_udf
  elif field["type"] == "double":
    ...

  if udf is not None:
    column_statement = udf(field["name"]).alias(field["name"])
  else:
    column_statement = field["name"]

  column_statements.append(column_statement)
```

grejdi-mbta: Yes, but I'm not mapping a string to a function. Code like `dict["func"]("arg")` always felt like it lacked clarity, and created an uncertainty of whether `"func"` actually was the function name or if it was named something else, creating one more hoop to jump through.

py_cubic_ingestion/lib/py_cubic_ingestion/custom_udfs.py
grejdi-mbta: Updated. Let me know what you think.

py_cubic_ingestion/lib/py_cubic_ingestion/ingest_incoming.py
```diff
@@ -3,14 +3,14 @@
 in `aws/s3/glue_jobs/cubic_ingestion/ingest_incoming.py`.
 """
 
-import os
```
grejdi-mbta, line 6: Reshuffling imports a little to make them alphabetical.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -75,6 +93,42 @@ def table_name_suffix(load_table_s3_prefix: str) -> str:
     return "__ct" if load_table_s3_prefix.endswith("__ct/") else ""
 
 
+def get_glue_table_schema_fields_by_load(glue_client: GlueClient, glue_database_name: str, load: dict) -> list:
```
grejdi-mbta, line 96: Note Glue Client is now being passed in. This helps with testing.

py_cubic_ingestion/lib/py_cubic_ingestion/custom_udfs.py
grejdi-mbta: Updated with other code.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
grejdi-mbta: Updated.

py_cubic_ingestion/lib/py_cubic_ingestion/custom_udfs.py
```diff
@@ -0,0 +1,65 @@
...
+    return wrapper
+
+
+@optional
+@capture_error
+def as_long(s: str) -> int:
+    # int is the type for long as well
+    val = int(s)
+    if val < -9223372036854775808 or val > 9223372036854775807:
+        raise ValueError(f"'{s}', Out of range of Spark LongType data type. Use DecimalType.")
```
paulswartz, line 40: question: Doesn't `@optional` handle the error case here?

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -109,6 +163,44 @@ def from_catalog_kwargs(load: dict, env: dict) -> dict:
...
+        # override if we can cast successfully
+        if field["type"] == "long":
+            column = as_long_udf(field["name"]).alias(field["name"])
+        elif field["type"] == "double":
+            column = as_double_udf(field["name"]).alias(field["name"])
+        elif field["type"] == "date":
+            column = as_date_udf(field["name"]).alias(field["name"])
+        elif field["type"] == "timestamp":
+            column = as_timestamp_udf(field["name"]).alias(field["name"])
+
```
paulswartz, line 198: suggestion(non-blocking): this still duplicates much of the functionality (calling the UDF, aliasing the column) and could be simplified.

py_cubic_ingestion/pyproject.toml
```diff
@@ -10,6 +10,8 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
+python-dateutil = "^2.8.2"
+boto3-stubs = {extras = ["glue"], version = "^1.24.23"}
```
paulswartz, line 14: praise: neat!

py_cubic_ingestion/tests/test_custom_udfs.py
```diff
@@ -0,0 +1,94 @@
...
+    assert custom_udfs.as_long("\v") is None
+    assert custom_udfs.as_long("\f") is None
+    assert custom_udfs.as_long("\r") is None
+    assert custom_udfs.as_long("1") is not None
+
+
+def test_as_long() -> None:
+    # empty
+    assert custom_udfs.as_long(None) is None
+    assert custom_udfs.as_long("") is None
```
paulswartz, line 28: question: doesn't this duplicate `test_as_long_optional`?

py_cubic_ingestion/lib/py_cubic_ingestion/custom_udfs.py
grejdi-mbta: No. It's captured by `@capture_error` decorator, which then adds on to the message.

py_cubic_ingestion/tests/test_custom_udfs.py
grejdi-mbta: Removed. I kept the other optional checks for `test_as_double` etc, in case the decorator is removed by accident.

py_cubic_ingestion/lib/py_cubic_ingestion/custom_udfs.py
paulswartz: Sorry, that's what I meant. If `@capture_error` is already including the value in the exception, do you need to also include it here?

grejdi-mbta: You're right. Updated. Thanks.

PR #65, opened 2022-06-30
[elixir+python] add raw for unprocessed springboard loads

grejdi-mbta:
This PR has code to support the implementation of this [ADR](https://github.com/mbta/data_platform/blob/main/doc/adr/0007-move-to-string-type-for-all-cubic-provided-data.md). Please take a look, before reviewing.

### DIFF
--- a/ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
+++ b/ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
@@ -4,7 +4,7 @@ defmodule ExCubicIngestion.Repo.Migrations.AddIsRawForTablesLoads do
   import Ecto.Query
 
   alias ExCubicIngestion.Repo
-  alias ExCubicIngestion.Schema.CubicLoad
+  alias ExCubicIngestion.Schema.CubicTable
 
   def up do
     alter table(:cubic_tables) do
@@ -17,7 +17,7 @@ defmodule ExCubicIngestion.Repo.Migrations.AddIsRawForTablesLoads do
 
     flush()
 
-    Repo.update_all(CubicLoad.not_deleted(), set: [is_raw: true])
+    Repo.update_all(CubicTable.not_deleted(), set: [is_raw: true])
   end
 
   def down do
--- a/ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
+++ b/ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
@@ -1,6 +1,11 @@
 defmodule ExCubicIngestion.Repo.Migrations.AddIsRawForTablesLoads do
   use Ecto.Migration
 
+  import Ecto.Query
+
+  alias ExCubicIngestion.Repo
+  alias ExCubicIngestion.Schema.CubicLoad
+
   def up do
     alter table(:cubic_tables) do
       add :is_raw, :boolean
@@ -9,6 +14,10 @@ defmodule ExCubicIngestion.Repo.Migrations.AddIsRawForTablesLoads do
     alter table(:cubic_loads) do
       add :is_raw, :boolean
     end
+
+    flush()
+
+    Repo.update_all(CubicLoad.not_deleted(), set: [is_raw: true])
   end
 
   def down do
--- a/ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
+++ b/ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
@@ -1,0 +1,23 @@
+defmodule ExCubicIngestion.Repo.Migrations.AddIsRawForTablesLoads do
+  use Ecto.Migration
+
+  def up do
+    alter table(:cubic_tables) do
+      add :is_raw, :boolean
+    end
+
+    alter table(:cubic_loads) do
+      add :is_raw, :boolean
+    end
+  end
+
+  def down do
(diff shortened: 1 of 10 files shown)


ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
```diff
@@ -0,0 +1,32 @@
...
+      add :is_raw, :boolean
+    end
+
+    alter table(:cubic_loads) do
+      add :is_raw, :boolean
+    end
+
+    flush()
+
+    Repo.update_all(CubicLoad.not_deleted(), set: [is_raw: true])
```
grejdi-mbta, line 20: We want current table to have this set to True.

grejdi-mbta:
> @mzappitello can you please review the python portion of this PR? @losvedir the Elixir side please? PR description points to documentation.

@losvedir @mzappitello ping! Please let me know if you are able to review.

ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
```diff
@@ -0,0 +1,32 @@
...
+      add :is_raw, :boolean
+    end
+
+    alter table(:cubic_loads) do
+      add :is_raw, :boolean
+    end
+
+    flush()
+
+    Repo.update_all(CubicTable.not_deleted(), set: [is_raw: true])
```
losvedir, line 20: I think it's better to copy in the `where: is_nil(table.deleted_at)` logic rather than using a model function here, even if it does lead to some code duplication.

The reason is that this command needs to run in the context of the schema as it is now, while `CubicTable.not_deleted()` will return a query based on the schema as it is in the current code. If, for example, in the future we change how "deleted" is managed (to a boolean or move to a "deleted" table or something like that), then that would retroactively break this migration, and a new developer, or the test suite, wouldn't be able to do an `mix ecto.setup`.

losvedir (approved):
Looks good to me. One question and an optional suggestion, but giving an optimistic approval.

ex_cubic_ingestion/priv/repo/migrations/20220629190702_add_is_raw_for_tables_loads.exs
```diff
@@ -0,0 +1,32 @@
...
+  use Ecto.Migration
+
+  import Ecto.Query
+
+  alias ExCubicIngestion.Repo
+  alias ExCubicIngestion.Schema.CubicTable
+
+  def up do
+    alter table(:cubic_tables) do
+      add :is_raw, :boolean
```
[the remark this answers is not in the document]
losvedir: Makes sense!

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -143,23 +118,33 @@ def from_catalog_kwargs(load: dict, env: dict) -> dict:
     Returns
     -------
     dict
-        Dictionary that will be passed in as kwargs
+        Dictionary that contains information
```
mzappitello, line 121: not exactly the most helpful comment.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -143,23 +118,33 @@ def from_catalog_kwargs(load: dict, env: dict) -> dict:
...
+    source_key = load_s3_key
+    destination_path = os.path.dirname(load_s3_key)
+
+    # for change tracking loads, add a suffix to the table names
+    if destination_path.endswith("__ct"):
+        source_table_name = f"{source_table_name}__ct"
+        destination_table_name = f"{destination_table_name}__ct"
+
+    # for raw loads, add a prefix to the table_name and adjust the destination path
+    if load["is_raw"]:
```
mzappitello, line 138: i don't have context for this, but i'm guessing that this could be an `elif` off of the above if block.

if thats not the case, you are setting `destination_table_name` in both if blocks so the order of these statements matters. probably worth a semi detailed comment if thats the case.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
```diff
@@ -143,23 +118,33 @@ def from_catalog_kwargs(load: dict, env: dict) -> dict:
...
-        "database": env["GLUE_DATABASE_INCOMING"],
-        "table_name": f'{load["table_name"]}{(data_catalog_table_name_suffix)}',
-        "additional_options": {
-            "paths": [f's3://{env["S3_BUCKET_INCOMING"]}/{env.get("S3_BUCKET_PREFIX_INCOMING", "")}{load_s3_key}']
-        },
-        "transformation_ctx": "table_df_read",
+        "source_table_name": source_table_name,
+        "destination_table_name": destination_table_name,
+        "source_key": f's3://{env["S3_BUCKET_INCOMING"]}/{env.get("S3_BUCKET_PREFIX_INCOMING", "")}{source_key}',
+        "destination_path": f's3a://{env["S3_BUCKET_SPRINGBOARD"]}/'
```
mzappitello, line 146: nitpick, but why is one of these a key and the other a path?

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
grejdi-mbta: Order doesn't matter here. It's just conditions that if met, then something needs to happen.

py_cubic_ingestion/lib/py_cubic_ingestion/job_helpers.py
grejdi-mbta: Ha! I'll update.

PR #66, opened 2022-07-14
[db] add all known DMAP feeds

grejdi-mbta:
This PR adds the currently available DMAP feeds. The Glue Catalog tables are add in this PR, which will need to be deployed first: https://github.com/mbta/devops/pull/500

### DIFF
--- a/ex_cubic_ingestion/priv/repo/migrations/20220711142139_add_dmap_feeds.exs
+++ b/ex_cubic_ingestion/priv/repo/migrations/20220711142139_add_dmap_feeds.exs
@@ -1,0 +1,101 @@
+defmodule ExCubicIngestion.Repo.Migrations.AddDmapFeeds do
+  use Ecto.Migration
+
+  alias ExCubicIngestion.Repo
+  alias ExCubicIngestion.Schema.CubicDmapFeed
+  alias ExCubicIngestion.Schema.CubicTable
+
+  @dmap_tables [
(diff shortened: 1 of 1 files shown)


ex_cubic_ingestion/priv/repo/migrations/20220711142139_add_dmap_feeds.exs
```diff
@@ -0,0 +1,101 @@
...
+
+  def down do
+    Repo.transaction(fn ->
+      Enum.each(@dmap_tables, fn dmap_table ->
+        Repo.delete!(CubicTable.get_by!(
+          name: dmap_table[:name]
+        ))
+
+        Repo.delete!(CubicDmapFeed.get_by!(
+          relative_url: dmap_table[:relative_url]
```
paulswartz, line 96: suggestion(non-blocking): you don't need the `Access` pattern with maps (here and elsewhere in the file):
```suggestion
          relative_url: dmap_table.relative_url
```

paulswartz (approved):
🍰

ex_cubic_ingestion/priv/repo/migrations/20220711142139_add_dmap_feeds.exs
grejdi-mbta: Updated. Will keep that in mind. Thanks.

PR #67, opened 2022-07-19
[elixir] feat: set pool size for dev.

grejdi-mbta:
In an attempt to narrow down why the ECS service keeps staring new containers, I'd like to make sure the a crash is not happening because of an aggressive number of connections are going over the limit. See for RDS limits: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Limits.html#RDS_Limits.MaxConnections

### DIFF
--- a/ex_cubic_ingestion/config/runtime.exs
+++ b/ex_cubic_ingestion/config/runtime.exs
@@ -6,7 +6,9 @@ config :ex_cubic_ingestion, ExCubicIngestion.Repo,
   hostname: System.get_env("DB_HOST"),
   password: System.get_env("DB_PASSWORD"),
   port: "DB_PORT" |> System.get_env("5432") |> String.to_integer(),
-  configure: {ExCubicIngestion.Repo, :before_connect, []}
+  configure: {ExCubicIngestion.Repo, :before_connect, []},
+  # default to 10
+  pool_size: "DB_POOL_SIZE" |> System.get_env("10") |> String.to_integer()
 
 # for testing, we'd like to change the database
 if config_env() == :test do
--- a/.env.template
+++ b/.env.template
@@ -5,6 +5,7 @@ DB_PORT=5432
 DB_NAME=data_platform
 DB_USER=postgres
 DB_PASSWORD=postgres
+DB_POOL_SIZE=10
 
 # buckets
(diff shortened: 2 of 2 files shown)


losvedir:
I'm a bit confused on how this app is setup. Here's my general understanding, based on how we usually have it at CTD.

There are three `mix` (Elixir) environments: `test`, which is what the test suite runs as, `dev` which is what is run locally, and `prod`, which is how we build releases that get deployed to AWS.

In AWS we typically have at least a `prod` and a `dev` environment, but those are all built via the `prod` *mix* environment, and are simply distinguished from each other via environment variables.

So given all that, I'm having trouble squaring fixing an ECS issue with a `dev.exs`. But I know that `data_platform` is a little fancy in its use of Docker, and might be different from all our other apps. Are you actually running it locally, but it's controlling something in AWS? Is it correct to change `dev.exs` like this after all?

Normally, what I would expect, if you want to change something in just the AWS dev environment, would be to add the configuration as an environment variable, and the `config` the application using `System.get_env(...)` in the `config/runtime.exs` file.

grejdi-mbta:
@losvedir Thanks for the comment. I have revised based on your suggestion. I always get tripped up with `dev` not actually being the 'dev' ECS environment. Coupled with the fact that we don't have a real `prod` environment yet, it gets confusing.

ex_cubic_ingestion/config/runtime.exs
```diff
@@ -6,7 +6,9 @@ config :ex_cubic_ingestion, ExCubicIngestion.Repo,
   hostname: System.get_env("DB_HOST"),
   password: System.get_env("DB_PASSWORD"),
   port: "DB_PORT" |> System.get_env("5432") |> String.to_integer(),
-  configure: {ExCubicIngestion.Repo, :before_connect, []}
+  configure: {ExCubicIngestion.Repo, :before_connect, []},
+  # default to 10
+  pool_size: "DB_POOL_SIZE" |> System.get_env("10") |> String.to_integer()
```
grejdi-mbta, line 11: Devops PR: https://github.com/mbta/devops/pull/484

losvedir (approved):

PR #72, opened 2022-08-02
[elixir] fix: start transaction when something will be inserted

grejdi-mbta:
This PR addresses an issue where a transaction is being started with nothing to do.


paulswartz (commented):
Do you need a transaction to ensure that you're reading a consistent set of objects from the DB?
And if you don't need a transaction for that, do you need a transaction at all? Or would a direct write work?

paulswartz (approved):
🍰

PR #62, opened 2022-06-20
[elixir] feat: make archive/error jobs more error proof

grejdi-mbta:
This PR modularizes the copy and delete functionality of the archive and error workers. It also adds checks for various states  prior to copying, to make adjustments to logic.

PR #73, opened 2022-08-08
[elixir] feat: adjust to querying loads with only recent s3 objects

grejdi-mbta:
This PR address an issues where a list of s3 objects that is checked against the `cubic_loads` table is rather large and therefore the query is very large. With this PR, we shrink the list down to only objects from the last 24 hours. The objects will have 24 hours to make it through the ingestion process. If the object is in the Incoming bucket for longer than 24 hours, then something has gone wrong.

In our case, the cause of this issue was the ECS task shutting down before the job was finished, and hence leaving loads in  the 'ingesting' status. According to [this](https://github.com/sorentwo/oban/issues/430) Oban should update its state and fail the jobs on a KILL command, but I suspect that the ECS task can't communicate with the DB due to overload.

Overall, this is an improvement to the process and keeps the pipeline steady. 

We will address loads stuck in 'ingesting' in a similar way to [this PR](https://github.com/mbta/data_platform/pull/62), which addresses loads stuck in 'archiving' and 'erroring'.

### DIFF
--- a/ex_cubic_ingestion/lib/ex_aws/helpers.ex
+++ b/ex_cubic_ingestion/lib/ex_aws/helpers.ex
@@ -3,8 +3,6 @@ defmodule ExAws.Helpers do
   Central place for useful functions that do ExAws work.
   """
 
-  require Logger
-
   @doc """
   Performs checks, before copying the source object to a destination object
   and deleting the source object (moving).
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
@@ -103,8 +103,8 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
   """
   @spec insert_from_object_with_table(map(), CubicTable.t()) :: t()
   def insert_from_object_with_table(object, table) do
-    last_modified = parse_and_drop_msec(object[:last_modified])
-    size = String.to_integer(object[:size])
+    last_modified = parse_and_drop_msec(object.last_modified)
+    size = String.to_integer(object.size)
 
     status =
       if size > 0 do
@@ -116,21 +116,24 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
     Repo.insert!(%__MODULE__{
       table_id: table.id,
       status: status,
-      s3_key: object[:key],
+      s3_key: object.key,
       s3_modified: last_modified,
       s3_size: size,
       is_raw: table.is_raw
     })
   end
 
+  @doc """
+  For a list of S3 objects, get all the possible matching load records.
+  """
   @spec get_by_objects(list()) :: [t()]
   def get_by_objects(objects) do
     # put together filters based on the object info
     filters =
       Enum.map(objects, fn object ->
-        last_modified = parse_and_drop_msec(object[:last_modified])
+        last_modified = parse_and_drop_msec(object.last_modified)
 
-        {object[:key], last_modified}
+        {object.key, last_modified}
       end)
 
     # we only want to query if we have filters because otherwise the query will the return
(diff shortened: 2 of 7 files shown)


ex_cubic_ingestion/lib/ex_aws/helpers.ex
```diff
@@ -3,8 +3,6 @@ defmodule ExAws.Helpers do
   Central place for useful functions that do ExAws work.
   """
 
-  require Logger
```
grejdi-mbta, line 6: Cleanup.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
```diff
@@ -103,8 +103,8 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
   """
   @spec insert_from_object_with_table(map(), CubicTable.t()) :: t()
   def insert_from_object_with_table(object, table) do
-    last_modified = parse_and_drop_msec(object[:last_modified])
-    size = String.to_integer(object[:size])
+    last_modified = parse_and_drop_msec(object.last_modified)
```
grejdi-mbta, line 106: Recommended adjustments.

ex_cubic_ingestion/test/ex_cubic_ingestion/schema/cubic_load_test.exs
```diff
@@ -19,11 +15,33 @@ defmodule ExCubicIngestion.Schema.CubicLoadTest do
         is_raw: false
       })
 
+    utc_now = DateTime.utc_now()
```
grejdi-mbta, line 18: Since we're dealing with only objects from the last 24 hours, we need to adjust tests accordingly.

paulswartz (commented):
as an alternate approach, did you consider chunking the objects passed into `insert_new_from_objects_with_table`?

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
```diff
@@ -25,9 +25,16 @@ defmodule ExCubicIngestion.Validators do
...
+  in the last 24 hours
+  """
   @spec valid_s3_object?(map()) :: boolean()
-  def valid_s3_object?(%{key: key, size: _size}) do
-    not String.ends_with?(key, "/")
+  def valid_s3_object?(%{key: key, last_modified: last_modified, size: _size}) do
+    {:ok, last_modified_dt, _offset} = DateTime.from_iso8601(last_modified)
+
+    not String.ends_with?(key, "/") and
+      DateTime.compare(last_modified_dt, DateTime.add(DateTime.utc_now(), -86_400)) == :gt
```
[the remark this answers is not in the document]
grejdi-mbta: I've separated them into 2 functions to make it clearer. When I used 2 Enum.filters, Credo was complaining about it, so I opted to use an anon function.

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
```diff
@@ -25,9 +25,26 @@ defmodule ExCubicIngestion.Validators do
...
+    false
+  end
+
+  @doc """
+  Only valid if the name ends with '.csv.gz' or '.csv' and has a size specified
+  """
   @spec valid_s3_object?(map()) :: boolean()
   def valid_s3_object?(%{key: key, size: _size}) do
-    not String.ends_with?(key, "/")
+    String.ends_with?(key, ".csv.gz") or String.ends_with?(key, ".csv")
```
grejdi-mbta, line 47: An update here to only look at CSV files, since more files have made it into the Incoming due to: https://app.asana.com/0/0/1202452782579116/1202779947051817/f

paulswartz:
> > as an alternate approach, did you consider chunking the objects passed into `insert_new_from_objects_with_table`?
> 
> Not sure what yo mean here. Where would the chunking occur?

From your description of the issue, my understanding is that `insert_new_from_objects_with_table` is checking too many S3 objects to insert into the database. The fix here is to limit the number of S3 objects we pass into the function, by filtering out any objects older than 24 hours.

My question is: rather than filtering out S3 objects, what if we made multiple calls to `insert_new_from_objects_with_table`?

```elixir
# process_incoming.ex
def run(state) do
  ...
  s3_objects
  |> Stream.chunk_by(1_000) # or some other limit
  |> Enum.flat_map(&insert_new_from_objects_with_table(&1, table))
end
```

paulswartz (approved):

PR #74, opened 2022-08-25
[elixir] feat: move DFM files out of incoming

grejdi-mbta:
Cubic has enabled a feature of Qlik Replicate that sends metadata files alongside data files. These metadata files contain the schema and basic information about the data, and are identified by the data files's name and the `.dfm` extension. In this PR, we move these files to the Archive and Error buckets, so they do not linger in Incoming. We also make some adjustments to help testing.


ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
```diff
@@ -40,11 +40,11 @@ defmodule ExCubicIngestion.Validators do
   end
 
   @doc """
-  Only valid if the name ends with '.csv.gz' or '.csv' and has a size specified
+  Only valid if the name ends with '.csv.gz' and has a size specified
   """
   @spec valid_s3_object?(map()) :: boolean()
   def valid_s3_object?(%{key: key, size: _size}) do
-    String.ends_with?(key, ".csv.gz") or String.ends_with?(key, ".csv")
```
grejdi-mbta, line 47: Removing support for `.csv` as it was complicating testing.

ex_cubic_ingestion/priv/repo/migrations/20220825021933_update_snapshot_key_for_sample.exs
```diff
@@ -0,0 +1,21 @@
...
+
+  alias ExCubicIngestion.Repo
+  alias ExCubicIngestion.Schema.CubicTable
+  alias ExCubicIngestion.Schema.CubicOdsTableSnapshot
+
+  def up do
+    table_rec = Repo.get_by!(CubicTable, name: "cubic_ods_qlik__edw_sample")
+    ods_table_snapshot_rec = Repo.get_by!(CubicOdsTableSnapshot, table_id: table_rec.id)
+
+    Repo.update!(Ecto.Changeset.change(ods_table_snapshot_rec, snapshot_s3_key: "cubic/ods_qlik/EDW.SAMPLE/LOAD1.csv.gz"))
```
grejdi-mbta, line 12: Migration for updating the snapshot key of the `EDW.SAMPLE` table.

ex_cubic_ingestion/priv/repo/migrations/20220825160222_add_sample_dmap_table.exs
```diff
@@ -0,0 +1,20 @@
+defmodule ExCubicIngestion.Repo.Migrations.AddSampleDmapTable do
+  use Ecto.Migration
+
+  alias ExCubicIngestion.Repo
+  alias ExCubicIngestion.Schema.CubicTable
+
+  def up do
+    Repo.insert!(%CubicTable{
```
grejdi-mbta, line 8: Migration for DMAP `agg_sample`, as I noticed I hadn't added it yet.

ex_cubic_ingestion/test/support/ex_aws_data.ex
```diff
@@ -15,31 +15,47 @@ defmodule MockExAws.Data do
...
-        key: "#{incoming_prefix}cubic/ods_qlik/SAMPLE/LOAD1.csv",
+        key: "#{incoming_prefix}cubic/ods_qlik/SAMPLE/LOAD1.csv.gz",
         last_modified: dt_adjust_and_format(utc_now, -2400),
         owner: nil,
         size: "197",
         storage_class: "STANDARD"
       },
+      %{
+        e_tag: "\"abc123\"",
+        key: "#{incoming_prefix}cubic/ods_qlik/SAMPLE/LOAD1.dfm",
```
grejdi-mbta, line 42: Added `.dfm` files.

paulswartz:
I thought this happened already? For example, `cubic/ods_qlik/EDW.CARD_DIMENSION/timestamp%3D20220810T201133Z/LOAD00000001.dfm` in the dev error bucket.

grejdi-mbta:
> I thought this happened already? For example, `cubic/ods_qlik/EDW.CARD_DIMENSION/timestamp%3D20220810T201133Z/LOAD00000001.dfm` in the dev error bucket.

Prior to restricting the files to `.csv.gz`, all these files that didn't conform were being routed to the error bucket as they were not recognized and didn't have data that made sense. Now, I'm being more intentional in their movement.

paulswartz (commented):
If there's nothing to do with these files at the moment, could you treat them as valid files, do no processing, and move them to the archive? That wouldn't require any updates to the archive or error workers (I think).

grejdi-mbta:
> If there's nothing to do with these files at the moment, could you treat them as valid files, do no processing, and move them to the archive? That wouldn't require any updates to the archive or error workers (I think).

So they are being treated as valid, and I'm not doing any processing on them. I'm just moving them to Archive/Error through the `archive` and `error` workers, as that's where the data file is being moved from as well.

paulswartz:
> So they are being treated as valid, and I'm not doing any processing on them. I'm just moving them to Archive/Error through the `archive` and `error` workers, as that's where the data file is being moved from as well.

But they're being special cased, which means each file the archive/error workers process is going to check for a `.dfm` file. How are the files getting picked up by the error worker now?

ex_cubic_ingestion/lib/ex_cubic_ingestion/validators.ex
```diff
@@ -40,11 +40,11 @@ defmodule ExCubicIngestion.Validators do
...
+    String.ends_with?(key, ".csv.gz")
```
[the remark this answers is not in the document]
grejdi-mbta: The naming convention for DFM files looks like this: For a data file, such as `LOAD001.csv.gz`, the metadata file is named `LOAD001.dfm`. I wanted to be able to just swap one extension for another, `.csv.gz` to `.dfm`. With the support of the `.csv`, I was adding extra logic that was only really supporting local test cases. So I standardized to `.csv.gz` for everything, and removed the logic to support `.csv` all around.

grejdi-mbta:
> But they're being special cased, which means each file the archive/error workers process is going to check for a `.dfm` file.
Only ODS is being checked for DFMs.
> How are the files getting picked up by the error worker now?
They had come into the database as load records, and because Glue couldn't process them, they were errored out.

paulswartz:
> They had come into the database as load records, and because Glue couldn't process them, they were errored out.

Would it be easier to update the Glue process to do nothing with `.dfm` files?

grejdi-mbta:
> Would it be easier to update the Glue process to do nothing with `.dfm` files?

Sorry, I guess I never hit submit on my comment. I was saying that I rather not. The Glue job has nothing to do with the DFM yet, so for now I rather not have it do anything with it.

paulswartz:
> > Would it be easier to update the Glue process to do nothing with `.dfm` files?
> 
> Sorry, I guess I never hit submit on my comment. I was saying that I rather not. The Glue job has nothing to do with the DFM yet, so for now I rather not have it do anything with it.

What about not including them in the request to Glue, then? https://github.com/mbta/data_platform/blob/9b23c1dbe0dbee2f210c0d277514fce15c48063c/ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex#L130-L141 builds the `loads` key which is passed to Glue: you could filter the DFM files from that key. The archive and error workers don't care what happened with the actual file, but it will see that the file is marked for archiving and handle it the same way it handles any other file.

Right now, it seems that every ODS file (load and change) is going to also check for a DFM file, which seems like overkill.

grejdi-mbta:
> What about not including them in the request to Glue, then?

Just want to speak this back to you. I think you're asking me if we have the DFM objects inserted into the `cubic_loads` table. When starting ingestion, we should ignore them from being sent to Glue, but at the end of the Oban job mark them as being `ready_for_archiving/ready_for_erroring` just like the data load record is currently. With this approach, we then don't need to make modification to the archive/error jobs and they will as if they are regular loads. Is this correct?

paulswartz:
That's correct. I think that would be a less invasive change, and you would only need to make the change in one place (Ingest worker) rather than two (both Archive and Error).

grejdi-mbta:
> That's correct. I think that would be a less invasive change, and you would only need to make the change in one place (Ingest worker) rather than two (both Archive and Error).

I agree that it's less invasive. It's just that inserting the DFMs in the database doesn't really represent them correctly. They are not load files. They are something else. I can see them being represented in a different way, such as within a `cubic_load_schemas` table with a `load_id` FK, but that feels like overkill. I like the approach in this PR because it's really doing that cleanup step. In a future where the `cubic_loads` are being observed through a dashboard, records that are for DFMs will stand out.

paulswartz (approved):
I don't think this is a good strategy. But I'm going on vacation and don't want to block this work any further.

Keep an eye on these two workers as they evolve: if they keep needing duplicate functionality (as they do here), you should think about pulling that out into a shared module.

grejdi-mbta:
> Keep an eye on these two workers as they evolve: if they keep needing duplicate functionality (as they do here), you should think about pulling that out into a shared module.

Yes. Of course. I know that this is a concern for you.

PR #75, opened 2022-08-25
[elixir] feat: limit how many ready loads to start ingestion for

grejdi-mbta:
This PR continues the work on optimizing the processes and creating a stable ECS container environment from this [Asana task](https://app.asana.com/0/1201701089427502/1202462260207371). We introduce a different way for querying for 'ready' loads by first getting all the active tables and then querying the 'ready' loads for each table (with a limit). For ODS, we also utilize the last snapshot load, so we only get the loads that we know exist and have not been remove by Qlik during a restart.  An [Asana ticket](https://app.asana.com/0/1201701089427502/1202900902122498) has been created to think through what we do with these 'ready' load records that are now pointing to non-existent objects.

Previously, if there were a lot of 'ready' loads, the process might take a really long time to get to the adding of the ingestion jobs.

### DIFF
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
@@ -136,15 +136,48 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
     })
   end
 
-  @spec get_status_ready :: [t()]
-  def get_status_ready do
-    query =
-      from(load in not_deleted(),
-        where: load.status == "ready",
-        order_by: [load.s3_modified, load.s3_key]
-      )
+  @doc """
+  Get loads with the 'ready' status by getting all the active tables and
+  querying for the first {limit} loads by table. For ODS, because there might be a
+  Qlik restart between the 'ready' and 'ingesting' status, we should check to make
+  sure we only start since the last snapshot load's s3_modified. Because of this
+  logic, some 'ready' loads will be left behind. That's because they will be
+  pointing to non-existing objects (Note: Qlik removes all objects on a restart).
+  """
+  @spec get_status_ready(integer()) :: [t()]
+  def get_status_ready(limit \\ 100) do
+    # we need to get 'ready' loads only for active tables
+    CubicTable.all_with_ods_table_snapshot()
+    |> Enum.map(fn {table_rec, ods_table_snapshot_rec} ->
+      # get the last load with snapshot key
+      last_snapshot_load_rec =
+        if not is_nil(ods_table_snapshot_rec) do
+          Repo.one(
+            from(load in not_deleted(),
+              where: load.s3_key == ^ods_table_snapshot_rec.snapshot_s3_key,
+              order_by: [desc: load.s3_modified],
+              limit: 1
+            )
+          )
+        end
 
-    Repo.all(query)
+      # query for 'ready' loads, with limit
+      ready_loads =
+        from(load in not_deleted(),
+          where: load.status == "ready" and load.table_id == ^table_rec.id,
+          order_by: [load.s3_modified, load.s3_key],
+          limit: ^limit
+        )
+
+      # if we know the last snapshot load, then filter down to 'ready' loads
+      # after the snapshot
+      if is_nil(last_snapshot_load_rec) do
+        ready_loads
+      else
+        from(load in ready_loads, where: load.s3_modified >= ^last_snapshot_load_rec.s3_modified)
+      end
+    end)
+    |> Enum.flat_map(&Repo.all(&1))
   end
 
   @doc """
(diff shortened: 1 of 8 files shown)


ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
```diff
@@ -136,15 +136,48 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
...
+            )
+          )
+        end
+
+      # query for 'ready' loads, with limit
+      ready_loads =
+        from(load in not_deleted(),
+          where: load.status == "ready" and load.table_id == ^table_rec.id,
+          order_by: [load.s3_modified, load.s3_key],
+          limit: ^limit
```
grejdi-mbta, line 169: Limit helps in the case that there is a lot of 'ready' loads.

ex_cubic_ingestion/lib/ex_cubic_ingestion/start_ingestion.ex
```diff
@@ -73,6 +73,8 @@ defmodule ExCubicIngestion.StartIngestion do
     ready_loads
     |> chunk_loads(@max_num_of_loads, @max_size_of_loads)
     |> Enum.each(&process_loads/1)
+
+    :ok
```
grejdi-mbta, line 77: Add this since it's in my other GenServers.

paulswartz: `Enum.each/2` already returns `:ok`, but it doesn't hurt to add.

ex_cubic_ingestion/test/ex_cubic_ingestion/schema/cubic_load_test.exs
```diff
@@ -125,6 +126,62 @@ defmodule ExCubicIngestion.Schema.CubicLoadTest do
...
+          status: "ready",
+          s3_key: "cubic/ods_qlik/SAMPLE/LOAD1.csv",
+          s3_modified: ~U[2022-01-03 20:49:50Z],
+          s3_size: 197,
+          is_raw: true
+        })
+
+      # ignoring loads prior to this last snapshot load
+      assert [new_ods_load] == CubicLoad.get_status_ready()
+    end
```
paulswartz, line 184: question: does this test represent how the `s3_key` values look in practice? I thought that the load files were always named "cubic/ods_qlik/<table>/LOAD0000001.csv.gz" but with a changed `s3_modified` value?

ex_cubic_ingestion/test/ex_cubic_ingestion/start_ingestion_test.exs
```diff
@@ -71,17 +19,29 @@ defmodule ExCubicIngestion.StartIngestionTest do
   end
 
   describe "run/0" do
-    test "schedules ingestion jobs for ready loads", %{
-      load_rec_ids: load_rec_ids
-    } do
+    test "schedules ingestion jobs for ready loads" do
```
paulswartz, line 22: question: why were the ODS load test cases removed?

paulswartz (commented):
It would be great to see some tests for a combination of the LOAD.csv.gz + `__ct` load files to see that they're working properly as well.

ex_cubic_ingestion/test/ex_cubic_ingestion/schema/cubic_load_test.exs
grejdi-mbta: You're right that the snapshot load files coming in now end with `LOAD0000001.csv.gz`. I didn't want to represent that in the tests as the code doesn't care about that specifically. It only cares about what the `snapshot_s3_key` is set in the database.

I did notice that these load files ended with `.csv`, so I fixed that to be consistent with the rest of the tests.

ex_cubic_ingestion/test/ex_cubic_ingestion/start_ingestion_test.exs
grejdi-mbta: I moved the setup inside the test, since it was only used once. I didn't think I needed the ODS load test, but upon a further look, I re-added because of the snapshot update that occurs within `StartIngestion.run`. The snapshot update is tested elsewhere, but it's good to capture the `filter` here. Thanks.

ex_cubic_ingestion/test/ex_cubic_ingestion/schema/cubic_load_test.exs
paulswartz: The reason I'm concerned is a situation like the following:
- `LOAD01.csv.gz` exists as a "ready" CubicLoad with `s3_modified` time X
- Qlik does a reload
- `LOAD01.csv.gz` still exists, but now there are two "ready" CubicLoads: one with an `s3_modified` time of X, and one with an `s3_modified` time of X + 1

grejdi-mbta: So the test captures this scenario. I noted in this PR's description that in the situation you are describing, the first load record will be ignored. I've added [an Asana task](https://app.asana.com/0/1201701089427502/1202900902122498) for trying to think through about what to do with these invalid 'ready' loads.

grejdi-mbta:
> It would be great to see some tests for a combination of the LOAD.csv.gz + `__ct` load files to see that they're working properly as well.

I forgot to address this comment. I have them for relevant areas, such as during insertion into the load table. Other than that, they are not treated any differently than regular loads.

PR #77, opened 2022-08-31
[elixir] feat: insert new objects only if new since last inserted load

grejdi-mbta:
This PR continues optimization work for creating a stable ECS container (see [Asana task](https://app.asana.com/0/1201701089427502/1202462260207371)). We introduce a new way of figuring out what is new from the list of S3 objects. We query for the last inserted load record for each table, and only insert S3 objects that are new since the last inserted load, using on the `s3_modified`.

### DIFF
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
@@ -74,18 +74,31 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
   end
 
   @doc """
-  From a list of S3 objects defined as maps, filter out any that have already
-  been inserted and therefore are being processed, and insert the rest in database.
+  From a list of S3 objects defined as maps, filter in only new objects and
+  insert them in database.
   """
   @spec insert_new_from_objects_with_table([map()], CubicTable.t()) ::
           {:ok, [t()]} | {:error, term()}
   def insert_new_from_objects_with_table(objects, table) do
-    # query loads to see what we can ignore when inserting
-    # usually happens when objects have not been moved out of 'incoming' bucket
-    recs = get_by_objects(objects)
+    # get the last inserted load which will be used to further filter the objects
+    last_inserted_load_rec =
+      Repo.one(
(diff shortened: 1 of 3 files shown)


ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
```diff
@@ -123,49 +136,6 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
     })
   end
 
-  @doc """
```
grejdi-mbta, line 126: No longer needed.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
```diff
@@ -74,18 +74,31 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
...
+  insert them in database.
   """
   @spec insert_new_from_objects_with_table([map()], CubicTable.t()) ::
           {:ok, [t()]} | {:error, term()}
   def insert_new_from_objects_with_table(objects, table) do
-    # query loads to see what we can ignore when inserting
-    # usually happens when objects have not been moved out of 'incoming' bucket
-    recs = get_by_objects(objects)
+    # get the last inserted load which will be used to further filter the objects
+    last_inserted_load_rec =
```
grejdi-mbta, line 84: Because ODS tables have 2 prefixes (batch and change tracking), we make this call twice for each ODS table for every run of this GenServer. There is a small optimization to cache the result. I've logged an Asana task for it.

paulswartz (commented):
I'm nervous that this will result in items being left behind -- were you able to test this in your environment?

grejdi-mbta:
> I'm nervous that this will result in items being left behind -- were you able to test this in your environment?

I have, and have also tested on Dev. My local does not push the limits of this code, so I've had to try it on Dev. It actually works fine, although I do have that nervousness as well. My nervousness stems from AWS's systems being scalable in nature and information not being represented through the API accurately at the time we request it. I'm going to experiment with moving date/times back to provide some buffers, but for now this works pretty well.

paulswartz (approved):

PR #82, opened 2022-09-23
[elixir] feat: add info for args when worker has exception

grejdi-mbta:
This PR adds more logging to help in narrowing down exceptions. We usually get this information when a Glue job starts and is running, but if the exception occurs before, we wouldn't have the information. We also add testing for the logs.

### DIFF
--- a/ex_cubic_ingestion/config/test.exs
+++ b/ex_cubic_ingestion/config/test.exs
@@ -1,7 +1,7 @@
 import Config
 
 # only log warnings+ in test
-config :logger, level: :warning
+config :logger, level: :info
 
 config :ex_cubic_ingestion,
   start_app_children?: false
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/oban_logger.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/oban_logger.ex
@@ -31,50 +31,37 @@ defmodule ExCubicIngestion.ObanLogger do
   def handle_event(
         [:oban, :job, :exception],
         measure,
-        %{worker: worker} = meta,
+        %{args: %{"load_rec_id" => load_rec_id}} = meta,
         _config
-      )
-      when worker in ["ExCubicIngestion.Workers.Archive", "ExCubicIngestion.Workers.Error"] do
-    %{"load_rec_id" => load_rec_id} = meta.args
-
-    args_info = CubicLoad.get!(load_rec_id)
-
-    log_exception(measure, meta, args_info)
+      ) do
+    log_exception(measure, meta, CubicLoad.get!(load_rec_id))
   end
 
   def handle_event(
         [:oban, :job, :exception],
         measure,
-        %{worker: "ExCubicIngestion.Workers.FetchDmap"} = meta,
+        %{args: %{"feed_id" => feed_id}} = meta,
         _config
       ) do
-    %{"feed_id" => feed_id} = meta.args
-
-    args_info = CubicDmapFeed.get!(feed_id)
-
-    log_exception(measure, meta, args_info)
+    log_exception(measure, meta, CubicDmapFeed.get!(feed_id))
   end
 
   def handle_event(
         [:oban, :job, :exception],
         measure,
-        %{worker: "ExCubicIngestion.Workers.Ingest"} = meta,
+        %{args: %{"load_rec_ids" => load_rec_ids}} = meta,
         _config
       ) do
-    %{"load_rec_ids" => load_rec_ids} = meta.args
-
-    args_info = CubicLoad.get_many_with_table(load_rec_ids)
-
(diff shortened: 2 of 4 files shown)


ex_cubic_ingestion/config/test.exs
```diff
@@ -1,7 +1,7 @@
 import Config
 
 # only log warnings+ in test
-config :logger, level: :warning
+config :logger, level: :info
```
grejdi-mbta, line 4: I'd like to capture more here due to the tests.

ex_cubic_ingestion/lib/ex_cubic_ingestion/oban_logger.ex
```diff
@@ -25,9 +28,53 @@ defmodule ExCubicIngestion.ObanLogger do
     )
   end
 
+  def handle_event(
+        [:oban, :job, :exception],
+        measure,
+        %{worker: worker} = meta,
+        _config
+      )
+      when worker in ["ExCubicIngestion.Workers.Archive", "ExCubicIngestion.Workers.Error"] do
```
grejdi-mbta, line 37: I think this is good, but let me know if there is a better way to do it.

ex_cubic_ingestion/lib/ex_cubic_ingestion/oban_logger.ex
```diff
@@ -25,9 +28,53 @@ defmodule ExCubicIngestion.ObanLogger do
...
+  end
+
   def handle_event([:oban, :job, :exception], measure, meta, _config) do
+    log_exception(measure, meta, {})
+  end
+
+  defp log_exception(measure, meta, args_info) do
     Logger.error(
-      "#{@log_prefix} [#{meta.queue}] Exception: args=#{inspect(meta.args, charlists: :as_lists)} duration=#{measure.duration} queue_time=#{measure.queue_time} state=#{meta.state} attempt=#{meta.attempt} kind=#{meta.kind} error=#{inspect(meta.error)}\nStacktrace:\n#{Exception.format_stacktrace(meta.stacktrace)}"
+      "#{@log_prefix} [#{meta.queue}] Exception: args=#{inspect(meta.args, charlists: :as_lists)} duration=#{measure.duration} queue_time=#{measure.queue_time} state=#{meta.state} attempt=#{meta.attempt} kind=#{meta.kind} error=#{inspect(meta.error)}\nargs_info: #{inspect(args_info)}\nStacktrace:\n#{Exception.format_stacktrace(meta.stacktrace)}"
```
grejdi-mbta, line 77: Using inspect here, and it dumps out the whole record. I think it's fine to dump this information in the log, but let me know what you think.

ex_cubic_ingestion/test/ex_cubic_ingestion/oban_logger_test.exs
```diff
@@ -0,0 +1,155 @@
...
+                   args: %{"load_rec_id" => dmap_load.id},
+                   state: "",
+                   attempt: 1,
+                   kind: "",
+                   error: "",
+                   stacktrace: nil
+                 },
+                 %{}
+               )
+             end) =~ dmap_load.s3_key
```
grejdi-mbta, line 80: The `s3_key` is really the most important piece of information I need.

ex_cubic_ingestion/lib/ex_cubic_ingestion/oban_logger.ex
paulswartz: suggestion(non-blocking): Not necessarily "better", but another approach would be to match on the data you need, rather than the worker:
```elixir
 def handle_event(
        [:oban, :job, :exception],
        measure,
        %{args: %{"load_rec_id" => load_rec_id}} = meta,
        _config
      )
```
This way you'll log exceptions for any worker where there's a load ID, not only the two that your currently have specified.

ex_cubic_ingestion/lib/ex_cubic_ingestion/oban_logger.ex
paulswartz: suggestion(non-blocking): By adding it on a new line, you're making it harder on yourself to line up the data in Splunk. Could you include it on the same line as you do with `meta.args`?

paulswartz (approved):
Some small notes, but nothing blocking.

ex_cubic_ingestion/lib/ex_cubic_ingestion/oban_logger.ex
grejdi-mbta: Thanks. Good tip, and updated.

ex_cubic_ingestion/lib/ex_cubic_ingestion/oban_logger.ex
grejdi-mbta: Done.

PR #83, opened 2022-09-28
[elixir] fix: processing of ODS loads without snapshot load

grejdi-mbta:


ex_cubic_ingestion/test/ex_cubic_ingestion/schema/cubic_load_test.exs
```diff
@@ -126,62 +126,6 @@ defmodule ExCubicIngestion.Schema.CubicLoadTest do
       # assert that the last record inserted comes back
       assert rest_new_load_recs == CubicLoad.get_status_ready()
     end
-
-    test "getting ODS 'ready' loads" do
```
grejdi-mbta, line 130: Moved these further down and adjusted them.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_load.ex
```diff
@@ -138,45 +139,14 @@ defmodule ExCubicIngestion.Schema.CubicLoad do
...
-      # if we know the last snapshot load, then filter down to 'ready' loads
-      # after the snapshot
-      if is_nil(last_snapshot_load_rec) do
-        ready_loads
-      else
-        from(load in ready_loads, where: load.s3_modified >= ^last_snapshot_load_rec.s3_modified)
-      end
-    end)
+    |> Enum.map(&get_status_ready_for_table_query(&1))
+    |> Enum.filter(&(!is_nil(&1)))
```
paulswartz, line 149: suggestion(non-blocking): rather than `map |> filter`, you can consider using `flat_map`. In this case, it look something like
```elixir
|> Enum.flat_map(&get_status_ready_for_table_queries/1)
...

def get_status_ready_for_table_queries(...) do
  if load_rec do
    [query]
  else
    []
  end
end
```

i.e. return an empty list in the `nil` case, and a list of one struct in the other cases.

paulswartz (commented):
We both had concerns about this change in #77. Given that it's experiencing issues that we can't reproduce in testing, should we consider reverting the change and trying a different approach to stabilize the container?

paulswartz (commented):
Were you able to reproduce the issue locally?

paulswartz:
If you can't reproduce the issue, and [everything is working](https://github.com/mbta/data_platform/pull/83#issuecomment-1262405463) with the smaller Qlik batches, do you still need to make this change?

grejdi-mbta:
> If you can't reproduce the issue, and [everything is working](https://github.com/mbta/data_platform/pull/83#issuecomment-1262405463) with the smaller Qlik batches, do you still need to make this change?

This code protects the integrity of the process further, so I'd like to deploy it.

paulswartz (approved):
🍰

PR #86, opened 2022-10-11
[elixir] feat: oban will wait for jobs to finish before shutting down

grejdi-mbta:
With this PR, we have configured Oban to wait until jobs have finished before allowing the shut down of the application, and hence the container.

### DIFF
--- a/ex_cubic_ingestion/config/config.exs
+++ b/ex_cubic_ingestion/config/config.exs
@@ -12,6 +12,8 @@ config :ex_cubic_ingestion,
 
 config :ex_cubic_ingestion, Oban,
   repo: ExCubicIngestion.Repo,
+  # 15 minutes
+  shutdown_grace_period: 900_000,
   plugins: [
     {
       Oban.Plugins.Cron,


ex_cubic_ingestion/config/config.exs
```diff
@@ -12,6 +12,8 @@ config :ex_cubic_ingestion,
 
 config :ex_cubic_ingestion, Oban,
   repo: ExCubicIngestion.Repo,
+  # 15 minutes
+  shutdown_grace_period: 900_000,
```
grejdi-mbta, line 16: This is in alignment with 'ingest' job timeout, and soon to be the container timeout. All other jobs, such as 'archive', 'error', 'fetch_dmap', or 'schedule_dmap', shouldn't take longer than 15 minutes.

grejdi-mbta:
@paulswartz This was driving me a little nuts, so I addressed it after doing some research with Oban's configuration.

ex_cubic_ingestion/config/config.exs
paulswartz: Is there also an `mbta/devops` PR to increase the timeout in the ECS task?

grejdi-mbta: There will be one. I'm in the process of moving the resources to the `dataplatform_dev` first, before I make a PR. I didn't want to do it before.

paulswartz: If you don't also make the change there, this change won't matter. Fargate will kill the after the `stopTimeout` even if the application hasn't stopped gracefully.

grejdi-mbta: So it seems the `stopTimeout` max is 120 seconds, so we won't be able to achieve a graceful shutdown. Looking into alternative ways of doing it.

paulswartz (commented):
Leaving a comment so that this drops off my review queue, but please re-request a review when it's ready!

grejdi-mbta:
Closing this out as can't find an alternative without manipulating Oban jobs directly, which I'd need to research how to do properly.

PR #87, opened 2022-10-20
[elixir] hotfix: get only one record for the ods load snapshot

grejdi-mbta:
This PR addresses an issue where while running an 'ingest' job, the app is pulling the snapshot record for the load, and it is receiving back thousands of records. It should really just receive one record back, but due to an issue in the start_ingestion process, these records kept inserting since the load was not being moved out of the 'ready' status.
There is some additional work to make sure this works correctly. They will be addressed as part of [this bug](https://app.asana.com/0/1201701089427502/1202701470464832). More specifically:
1. Need to change 'start_ingestion' to just run through and update ODS table snapshot, insert load snapshot, check schema, and update status of load to 'ready_for_ingesting', all within a transaction.
2. Update 'process_ingestion' to also start 'ingest' jobs, based on 'ready_for_ingesting' status.

### DIFF
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_ods_load_snapshot.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_ods_load_snapshot.ex
@@ -49,6 +49,17 @@ defmodule ExCubicIngestion.Schema.CubicOdsLoadSnapshot do
     Repo.get_by!(not_deleted(), clauses, opts)
   end
 
+  @spec get_latest_by!(Keyword.t() | map()) :: t() | nil
+  def get_latest_by!(clauses) do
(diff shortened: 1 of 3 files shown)


paulswartz (approved):
:cake:

This seems fine. But, I'm concerned about all the trouble we seem to be having keeping the Load and Snapshot data in sync. Maybe it's worth taking a step back, to think about whether this approach is working?

grejdi-mbta:
> 🍰
> 
> This seems fine. But, I'm concerned about all the trouble we seem to be having keeping the Load and Snapshot data in sync. Maybe it's worth taking a step back, to think about whether this approach is working?

Hey, not ignoring you here. I agree, and I'm thinking through it. I'm trying a few approaches to see how they feel/look, and hoping to have something more concrete by the end of the week.

PR #88, opened 2022-10-20
[elixir] fix: ehmon config in the correct place

grejdi-mbta:
This PR addresses a warning on local for not having 'ehmon' as a dependency when running the application. Moving it to the prod configuration avoids this warning and is the correct place to put this.


paulswartz (approved):
🍰 

The warning is because you specified `:ehmon` as a prod-only dependency, so you shouldn't configure it in other environments. The other option would have been to change `mix.exs` so that `:ehmon` is available in all environments.

PR #94, opened 2023-02-03
[data] fix: add PK info for dfm sample data

grejdi-mbta:
Solved the mystery for how we can update rows. Qlik sends over a primary key designation for tables through their `.dfm` files. We will eventually use this to make targeted updates to rows at a final destination beyond the data lake. In this PR, we wanted to update the sample data to show this designation.

### DIFF
--- a/sample_data/cubic/ods_qlik/EDW.SAMPLE__ct/20211201-112233444.dfm
+++ b/sample_data/cubic/ods_qlik/EDW.SAMPLE__ct/20211201-112233444.dfm
@@ -1,97 +1,105 @@
 {
-    "dfmVersion":   "1.1",
+    "dfmVersion": "1.1",
     "taskInfo": {
         "name": "",
-        "sourceEndpoint":   "",
-        "sourceEndpointType":   "",
-        "sourceEndpointUser":   "",
-        "replicationServer":    "",
-        "operation":    ""
+        "sourceEndpoint": "",
+        "sourceEndpointType": "",
+        "sourceEndpointUser": "",
+        "replicationServer": "",
+        "operation": ""
     },
     "fileInfo": {
         "name": "20211201-112233444.csv",
-        "extension":    "gz",
+        "extension": "gz",
         "location": "incoming/cubic/ods_qlik/EDW.SAMPLE__ct",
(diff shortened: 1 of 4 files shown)


sample_data/cubic/ods_qlik/EDW.SAMPLE__ct/20211201-112233444.dfm
```diff
@@ -1,97 +1,105 @@
...
-                "length":   10,
-                "precision":    0,
-                "scale":    0,
-                "primaryKeyPos":    0
-            }, {
-                "ordinal":  5,
+                "length": 10,
+                "precision": 0,
+                "scale": 0,
+                "primaryKeyPos": 1
```
grejdi-mbta, line 74: Updated

sample_data/cubic/ods_qlik/EDW.SAMPLE__ct/20211201-122433444.dfm
```diff
@@ -1,97 +1,105 @@
...
-                "length":   10,
-                "precision":    0,
-                "scale":    0,
-                "primaryKeyPos":    0
-            }, {
-                "ordinal":  5,
+                "length": 10,
+                "precision": 0,
+                "scale": 0,
+                "primaryKeyPos": 1
```
grejdi-mbta, line 74: Updated. Everything else is format changes.

paulswartz (commented):
Are we receiving data with the `primaryKeyPos` boolean, and the sample data was wrong in this case? Or is this a feature that Cubic needs to enable in Qlik?

grejdi-mbta:
> Are we receiving data with the `primaryKeyPos` boolean, and the sample data was wrong in this case? Or is this a feature that Cubic needs to enable in Qlik?

Sample data was wrong, and it's actually not a boolean. Qlik has this information currently. For example, a table in ODS might have a primary key comprised of two columns "tap_id" and "source". The Qlik DFM file will indicate that those columns have `"primaryKeyPos": 1` and `"primaryKeyPos": 2` set respectively.

paulswartz (approved):

PR #95, opened 2023-02-11
[elixir] fix: ingestion process crashes when missing schema file

grejdi-mbta:
This PR addresses an issue where a missing schema file for a data file was causing the validate_incoming process to crash. We now log this issue and allow the process to move to the load to a 'ready_for_erroring' status.

### DIFF
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema_fetch.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema_fetch.ex
@@ -6,6 +6,8 @@ defmodule ExCubicIngestion.SchemaFetch do
   alias ExCubicIngestion.Schema.CubicLoad
   alias ExCubicIngestion.Schema.CubicTable
 
+  require Logger
+
   @spec get_cubic_ods_qlik_columns(module(), CubicLoad.t()) :: list()
   @doc """
   Using the load's S3 key, we identify the Cubic provided schema and extract the columns.
@@ -18,11 +20,8 @@ defmodule ExCubicIngestion.SchemaFetch do
 
     incoming_bucket
     |> ExAws.S3.get_object("#{incoming_prefix}#{s3_key_root}.dfm")
-    |> lib_ex_aws.request!()
-    |> Map.fetch!(:body)
-    |> Jason.decode!()
-    |> map_fetch_cubic_ods_columns()
-    |> Enum.map(&String.downcase(Map.fetch!(&1, "name")))
+    |> lib_ex_aws.request()
+    |> handle_get_cubic_ods_columns()
   end
 
   @spec get_glue_columns(module(), CubicTable.t(), CubicLoad.t()) :: list()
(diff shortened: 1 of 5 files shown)


ex_cubic_ingestion/lib/ex_cubic_ingestion/validate_incoming.ex
```diff
@@ -84,8 +84,11 @@ defmodule ExCubicIngestion.ValidateIncoming do
   defp valid_schema?(state, {load_rec, table_rec}) do
     # if ODS, check the schema provided by Cubic (.dfm files) against Glue
     if CubicLoad.ods_load?(load_rec.s3_key) do
-      SchemaFetch.get_cubic_ods_qlik_columns(state.lib_ex_aws, load_rec) ==
-        SchemaFetch.get_glue_columns(state.lib_ex_aws, table_rec, load_rec)
+      qlik_columns = SchemaFetch.get_cubic_ods_qlik_columns(state.lib_ex_aws, load_rec)
+
+      not Enum.empty?(qlik_columns) and
```
paulswartz, line 89: question: is there a reason why you need to check for empty columns here? I would think that is another form of not-equals.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema_fetch.ex
```diff
@@ -46,6 +45,23 @@ defmodule ExCubicIngestion.SchemaFetch do
...
+    |> map_fetch_cubic_ods_columns()
+    |> Enum.map(&String.downcase(Map.fetch!(&1, "name")))
+  end
+
+  defp handle_get_cubic_ods_columns({:error, response}) do
+    Logger.error(
+      "[ex_cubic_ingestion] [schema_fetch] Unable to fetch schema: #{inspect(response)}"
+    )
+
+    []
```
paulswartz, line 62: suggestion: a more Elixir-y way of doing this would be to have get_cubic_ods_qlik_columns return an `{:ok, columns}` or `{:error, reason}` tuple. This would allow clients to distinguish between a table with no columns and an error.

ex_cubic_ingestion/lib/ex_cubic_ingestion/validate_incoming.ex
grejdi-mbta: I was trying to account for empty columns on both qlik schema and glue catalog. This was I'm being specific in that if we don't Qlik columns, then we essentially have no schema.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema_fetch.ex
grejdi-mbta: Thought about that. I just didn't want to pass the logic up and into validate_incoming.ex.

ex_cubic_ingestion/lib/ex_cubic_ingestion/validate_incoming.ex
paulswartz: But if there are no columns in Glue Catalog, shouldn't there also be no columns in the DFM?

grejdi-mbta: That shouldn't happen. We don't create tables in the Glue catalog without columns. It just wouldn't make sense.

paulswartz: I don't understand your previous comment then. If there are always columns in the Glue Catalog, then it's redundant to check whether the Qlik columns are empty?

grejdi-mbta: Just trying to be safe, as I can't guarantee how tables will be created in the future. Do you want me to remove this check?

grejdi-mbta: Removed.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema_fetch.ex
grejdi-mbta: Resolved this by removing extra check in validate_incoming.

PR #97, opened 2023-05-19
[elixir] feat: check DMAP config values before running DMAP fetch process

napilla:
Asana ticket: [Put DMAP fetch process behind configuration flag, so we can deploy without it.](https://app.asana.com/0/0/1204592266198521/f)

Was able to run the existing tests with the DMAP config values defined and they passed - with the values empty I can see the following is logged:
`10:04:20.543 [error] [ex_cubic_ingestion] [workers] [schedule_dmap] Exception: args=%{} duration=75000 queue_time=-6000 state=failure attempt=1 kind=error error=%Oban.PerformError{message: "ExCubicIngestion.Workers.ScheduleDmap failed with {:error, \"dmap_base_url or dmap_api_key is empty, not running dmap job\"}", reason: {:error, "dmap_base_url or dmap_api_key is empty, not running dmap job"}} args_info=nil`

### DIFF
--- a/ex_cubic_ingestion/config/config.exs
+++ b/ex_cubic_ingestion/config/config.exs
@@ -16,7 +16,7 @@ config :ex_cubic_ingestion, Oban,
     {
       Oban.Plugins.Cron,
       crontab: [
-        {"0 15 * * *", ExCubicIngestion.Workers.ScheduleDmap, max_attempts: 1}
+        {"* * * * *", ExCubicIngestion.Workers.ScheduleDmap, max_attempts: 1}
       ]
     }
   ],
--- a/ex_cubic_ingestion/config/config.exs
+++ b/ex_cubic_ingestion/config/config.exs
@@ -16,7 +16,7 @@ config :ex_cubic_ingestion, Oban,
     {
       Oban.Plugins.Cron,
       crontab: [
-        {"* * * * *", ExCubicIngestion.Workers.ScheduleDmap, max_attempts: 1}
+        {"0 15 * * *", ExCubicIngestion.Workers.ScheduleDmap, max_attempts: 1}
       ]
     }
   ],
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/application.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/application.ex
@@ -12,7 +12,7 @@ defmodule ExCubicIngestion.Application do
       {ExCubicIngestion.Repo.Migrator,
        run_migrations_at_startup?:
          Application.get_env(:ex_cubic_ingestion, :run_migrations_at_startup?)},
-      {Oban, ReleaseTasks.ScheduleDmap.oban_config()}
+      {Oban, Application.fetch_env!(:ex_cubic_ingestion, Oban)}
     ]
 
     app_children =
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/application.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/application.ex
@@ -12,7 +12,7 @@ defmodule ExCubicIngestion.Application do
       {ExCubicIngestion.Repo.Migrator,
        run_migrations_at_startup?:
          Application.get_env(:ex_cubic_ingestion, :run_migrations_at_startup?)},
-      {Oban, Application.fetch_env!(:ex_cubic_ingestion, Oban)}
+      {Oban, ReleaseTasks.ScheduleDmap.oban_config()}
     ]
 
     app_children =
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/schedule_dmap.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/schedule_dmap.ex
@@ -17,20 +17,10 @@ defmodule ExCubicIngestion.Workers.ScheduleDmap do
 
   @impl Oban.Worker
   def perform(_job) do
-    dmap_base_url = Application.fetch_env!(:ex_cubic_ingestion, :dmap_base_url)
-    dmap_api_key = Application.fetch_env!(:ex_cubic_ingestion, :dmap_api_key)
+    Repo.transaction(fn ->
+      Enum.each(CubicDmapFeed.all(), &Oban.insert(FetchDmap.new(%{feed_id: &1.id})))
+    end)
 
-    if dmap_base_url == "" or dmap_api_key == "" do
-      {:error, "dmap_base_url or dmap_api_key is empty, not running dmap job"}
-    else
-      Repo.transaction(fn ->
-        Enum.each(CubicDmapFeed.all(), &Oban.insert(FetchDmap.new(%{feed_id: &1.id})))
-      end)
-
-      :ok
-    end
-  rescue
-    e in ArgumentError ->
-      {:error, "dmap_base_url or dmap_api_key undefined, not running dmap job: " <> e.message}
+    :ok
   end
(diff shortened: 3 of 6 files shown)


ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/schedule_dmap.ex
```diff
@@ -17,10 +17,22 @@ defmodule ExCubicIngestion.Workers.ScheduleDmap do
 
   @impl Oban.Worker
   def perform(_job) do
-    Repo.transaction(fn ->
-      Enum.each(CubicDmapFeed.all(), &Oban.insert(FetchDmap.new(%{feed_id: &1.id})))
-    end)
+    try do
```
paulswartz, line 20: question: what is triggering this job? can we update that to not schedule the job at all if the DMAP environment is not configured?

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/schedule_dmap.ex
```diff
@@ -17,10 +17,20 @@ defmodule ExCubicIngestion.Workers.ScheduleDmap do
 
   @impl Oban.Worker
   def perform(_job) do
-    Repo.transaction(fn ->
-      Enum.each(CubicDmapFeed.all(), &Oban.insert(FetchDmap.new(%{feed_id: &1.id})))
-    end)
+    dmap_base_url = Application.fetch_env!(:ex_cubic_ingestion, :dmap_base_url)
```
grejdi-mbta, line 20: @napilla You'll need to update tests for this module.

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/schedule_dmap.ex
```diff
@@ -17,10 +17,20 @@ defmodule ExCubicIngestion.Workers.ScheduleDmap do
...
+    dmap_api_key = Application.fetch_env!(:ex_cubic_ingestion, :dmap_api_key)
 
-    :ok
+    if dmap_base_url == "" or dmap_api_key == "" do
```
grejdi-mbta, line 23: @napilla In an effort to work on you elixir skills, there is probably an opportunity to us pattern matching here.

grejdi-mbta (changes requested):

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/schedule_dmap.ex
napilla: Since I'm doing the check elsewhere now (when grabbing Oban config item in application startup) I added another test that checks this method in `ScheduleDmapTest`

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/schedule_dmap.ex
napilla: Thanks, I learned something here - let me know if my approach makes sense now

ex_cubic_ingestion/lib/release_tasks/schedule_dmap.ex
```diff
@@ -20,4 +22,32 @@ defmodule ReleaseTasks.ScheduleDmap do
...
+      ]
+      |> Enum.map(fn
+        :error -> false
+        {:ok, ""} -> false
+        _ok -> true
+      end)
+      |> Enum.all?(fn x -> x end)
+
+    case dmap_is_configured do
+      true ->
```
paulswartz, line 41: `case` with only booleans is clearer as an `if`. I think what @grejdi-mbta meant is something like:

```elixir
maybe_base_url = Application.fetch_env(:ex_cubic_ingestion, :dmap_base_url)
maybe_api_key = Application.fetch_env(:ex_cubic_ingestion, :dmap_base_url)

case {maybe_base_url, maybe_api_key} do
  {{:ok, base_url}, {:ok, api_key}} when base_url != "" and api_key != "" ->
    # configured
  _ ->
    # not configured
end
```

ex_cubic_ingestion/lib/release_tasks/schedule_dmap.ex
```diff
@@ -20,4 +22,32 @@ defmodule ReleaseTasks.ScheduleDmap do
...
+        Application.fetch_env!(:ex_cubic_ingestion, Oban)
+
+      false ->
+        Logger.warning("dmap_base_url or dmap_api_key empty, dmap will not be scheduled")
+
+        Enum.map(Application.fetch_env!(:ex_cubic_ingestion, Oban), fn
+          {:plugins, _} -> {:plugins, false}
```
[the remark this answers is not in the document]
napilla: I suppose none of them will be scheduled, if the dmap job isn't configured by then :) 

Good point, I'm wondering if it makes sense to define something that would build up the plugins (such as, we define the plugins as separate config items, and can add them into Oban as needed) that way whether or not we want to include this `ScheduleDmap` plugin can affect other plugins.

ex_cubic_ingestion/lib/ex_cubic_ingestion/application.ex
```diff
@@ -12,7 +12,7 @@ defmodule ExCubicIngestion.Application do
       {ExCubicIngestion.Repo.Migrator,
        run_migrations_at_startup?:
          Application.get_env(:ex_cubic_ingestion, :run_migrations_at_startup?)},
-      {Oban, Application.fetch_env!(:ex_cubic_ingestion, Oban)}
+      {Oban, ReleaseTasks.ScheduleDmap.oban_config()}
```
grejdi-mbta, line 15: @napilla I feel like the other approach was better, just needed a little more cleanup (see Paul's comment). This structure assumes that Oban is only used for scheduling and that's not the truth. It actually allows for other jobs to be maintained. Lastly, per the AC in the Asana ticket, I'd like to have this logged every day that it's not running, not just when the app starts.

grejdi-mbta (changes requested):

ex_cubic_ingestion/lib/release_tasks/schedule_dmap.ex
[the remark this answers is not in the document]
napilla: Or, given this approach has implications on how we run other plugins, and to meet the ACs (see @grejdi-mbta's comment) it seems best to move forward with the check for the config values in `schedule_dmap.ex`

ex_cubic_ingestion/lib/ex_cubic_ingestion/application.ex
napilla: Thanks for the clarity, updated here. I updated the existing tests to check that the job wouldn't run as expected as well.

ex_cubic_ingestion/config/config.exs
```diff
@@ -16,7 +16,7 @@ config :ex_cubic_ingestion, Oban,
     {
       Oban.Plugins.Cron,
       crontab: [
-        {"0 15 * * *", ExCubicIngestion.Workers.ScheduleDmap, max_attempts: 1}
```
grejdi-mbta, line 19: @napilla Please revert this change. Might have been for you to test I believe.

grejdi-mbta (changes requested):

ex_cubic_ingestion/config/config.exs
napilla: Yes, was testing the job was running. My apologies and fixed in latest push

grejdi-mbta (approved):

PR #100, opened 2023-09-21
[env+ga] fix: fix versions for building local and on github

grejdi-mbta:
Address two issues:
* Glue build was not working, hence we couldn't deploy.
* Local build was failing due to python3.7 version update and a poetry issue with the latest version.

### DIFF
--- a/.github/workflows/glue-python-deploy.yaml
+++ b/.github/workflows/glue-python-deploy.yaml
@@ -58,7 +58,7 @@ jobs:
       - name: Prepare for installing dependencies and package
         working-directory: ${{github.workspace}}/py_cubic_ingestion
         run: |
-          poetry export --output requirements.txt
+          poetry export --without-hashes --output requirements.txt
           poetry build
       - name: Install dependencies and py_cubic_ingestion package into 'libs'
         working-directory: ${{github.workspace}}/py_cubic_ingestion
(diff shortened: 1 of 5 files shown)


.github/workflows/glue-python-deploy.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
       - name: Prepare for installing dependencies and package
         working-directory: ${{github.workspace}}/py_cubic_ingestion
         run: |
-          poetry export --output requirements.txt
+          poetry export --without-hashes --output requirements.txt
```
grejdi-mbta, line 61: Addressing this issue: https://github.com/python-poetry/poetry/issues/3472

docker/Dockerfile.glue_2_0__local
```diff
@@ -31,7 +31,7 @@ ENV SPARK_HOME=/spark-2.4.3-bin-spark-2.4.3-bin-hadoop2.8
 RUN /glue/bin/gluesparksubmit --version
 
 # python packaging
-RUN pip install poetry
+RUN pip install poetry==1.3.2
```
grejdi-mbta, line 34: Addressing this issue: https://github.com/python-poetry/poetry/issues/7611

napilla: Would it make sense to comment here that the asdf tools version should be updated as well, if the version here needs to be updated in the future?

napilla (approved):
LGTM, approving and wonder if we need to document where versions should be changed in the project, if we need to update `poetry` versions again.

docker/Dockerfile.glue_2_0__local
grejdi-mbta: Added note, and point taken. I will make an attempt to ensure docs reflect this requirement.

PR #101, opened 2023-09-21
feat: use AWS IAM role instead of IAM user

ianwestcott:


krisrjohnson21 (approved):
🧼

krisrjohnson21:
Non-blocking question: Do we need a `role-duration-seconds` value, per the [original article](https://benoitboure.com/securely-access-your-aws-resources-from-github-actions) that captures this work?

krisrjohnson21:
> Non-blocking question: Do we need a `role-duration-seconds` value, per the [original article](https://benoitboure.com/securely-access-your-aws-resources-from-github-actions) that captures this work?

Looking at [build-push-ecr](https://github.com/mbta/actions/blob/2bfcdf225415d25a4a80e605afdbfcc38d0ab2c3/build-push-ecr/action.yml#L33), my guess is "no"

grejdi-mbta (approved):

PR #92, opened 2022-12-13
[elixir] feat: add athena partition after successful glue job

grejdi-mbta:

PR #102, opened 2023-09-27
[elixir] feat: scheduled loading of partitions in athena

grejdi-mbta:
Note: This work is related to the deployment of the data pipeline to Production.
The system will need to refresh partitions for some tables in Athena, so we are scheduling a job to run every hour to do so. The reason for the refresh is so our queries will return data under those partitions.
Without this job, we'd have to do something more manual. Not to mention that we'd need permissions to do it against live data.

**This work is dependent on a DevOps PR (forthcoming) to allow permissions for ECS to access Athena.**

### DIFF
--- a/ex_cubic_ingestion/config/config.exs
+++ b/ex_cubic_ingestion/config/config.exs
@@ -16,7 +16,17 @@ config :ex_cubic_ingestion, Oban,
     {
       Oban.Plugins.Cron,
       crontab: [
-        {"0 15 * * *", ExCubicIngestion.Workers.ScheduleDmap, max_attempts: 1}
+        {"0 15 * * *", ExCubicIngestion.Workers.ScheduleDmap, max_attempts: 1},
+        {"0 * * * *", ExCubicIngestion.Workers.LoadPartitions,
+         args: %{
+           table_names: [
+             "raw_cubic_ods_qlik__edw_fnp_general_jrnl_account_entry",
+             "raw_cubic_ods_qlik__edw_fnp_general_jrnl_account_entry__ct",
+             "raw_cubic_ods_qlik__edw_cch_afc_transaction",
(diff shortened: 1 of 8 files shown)


grejdi-mbta:
@paulswartz please review when you get the chance. I'd like to finalize this while I work on a DevOps PR to get Athena access for the container.

paulswartz (commented):
question: since this application is also writing the data into the partitions, could it trigger the repair when writing files? (for example, using [`ALTER TABLE ADD PARTITION`](https://docs.aws.amazon.com/athena/latest/ug/alter-table-add-partition.html))

grejdi-mbta:
> question: since this application is also writing the data into the partitions, could it trigger the repair when writing files? (for example, using [`ALTER TABLE ADD PARTITION`](https://docs.aws.amazon.com/athena/latest/ug/alter-table-add-partition.html))

Yes. Doing that kind of work in this PR: https://github.com/mbta/data_platform/pull/92, but I need to get something up quickly and more targeted for now to get through some of the FMIS tests that are planned next week.

paulswartz:
Could you run the commands manually for the test? Given that the previous work in #92 appears to have stalled for 8 months, I'm a little worried that we'll put in this "temporary" solution and it'll be a while before we come back to address it again.

grejdi-mbta:
> Could you run the commands manually for the test?

The "test" is running on production. I'd have to run these queries on live which I don't want to do, since it would give me access to live data.

> Given that the previous work in #92 appears to have stalled for 8 months, I'm a little worried that we'll put in this "temporary" solution and it'll be a while before we come back to address it again.

Agree with this. I'm trying to buy a little time for the team, while they find their groove with the current work. We will address for sure.

paulswartz (approved):

PR #104, opened 2023-11-01
[elixir] fix: disable load partitions for now

grejdi-mbta:
With the recent changes to permissions, we are able to run these queries manually for production. With this PR, we'll disable them to allow for a better solution through #92

### DIFF
--- a/ex_cubic_ingestion/config/config.exs
+++ b/ex_cubic_ingestion/config/config.exs
@@ -16,17 +16,17 @@ config :ex_cubic_ingestion, Oban,
     {
       Oban.Plugins.Cron,
       crontab: [
-        {"0 15 * * *", ExCubicIngestion.Workers.ScheduleDmap, max_attempts: 1},
-        {"0 * * * *", ExCubicIngestion.Workers.LoadPartitions,
-         args: %{
-           table_names: [
-             "raw_cubic_ods_qlik__edw_fnp_general_jrnl_account_entry",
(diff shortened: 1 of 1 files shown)


ex_cubic_ingestion/config/config.exs
```diff
@@ -16,17 +16,17 @@ config :ex_cubic_ingestion, Oban,
...
+        # {"0 * * * *", ExCubicIngestion.Workers.LoadPartitions,
+        #  args: %{
+        #    table_names: [
+        #      "raw_cubic_ods_qlik__edw_fnp_general_jrnl_account_entry",
+        #      "raw_cubic_ods_qlik__edw_fnp_general_jrnl_account_entry__ct",
+        #      "raw_cubic_ods_qlik__edw_cch_afc_transaction",
+        #      "raw_cubic_ods_qlik__edw_cch_afc_transaction__ct"
+        #    ]
+        #  },
+        #  max_attempts: 1}
```
paulswartz, line 29:
```suggestion
```

suggestion(non-blocking): you don't need to commit commented-out code. It's in the Git history if we need it again.

paulswartz (approved):

ex_cubic_ingestion/config/config.exs
grejdi-mbta: We'll remove later. Since if I remove the comment, I'd have to clean up `ExCubicIngestion.Workers.LoadPartitions`, and I'm not ready to do that just yet.

PR #112, opened 2023-12-29
[elixir] chore: update elixir mix versions

grejdi-mbta:


grejdi-mbta:
Updating now with #124.

PR #113, opened 2024-01-02
[elixir] feat: add is_active to cubic tables and filter not active from ingestion

grejdi-mbta:
I'm planning on adding an "Admin" area to manage the "active" status of tables. I will continue to manage migrations in `ex_cubic_ingestion`, so hence this change.
This PR also introduced the filtering by active on incoming loads.

### DIFF
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_table.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_table.ex
@@ -16,6 +16,7 @@ defmodule ExCubicIngestion.Schema.CubicTable do
              :name,
              :s3_prefix,
              :is_raw,
+             :is_active,
              :deleted_at,
              :inserted_at,
              :updated_at
@@ -26,6 +27,7 @@ defmodule ExCubicIngestion.Schema.CubicTable do
           name: String.t() | nil,
           s3_prefix: String.t() | nil,
           is_raw: boolean() | nil,
+          is_active: boolean() | nil,
           deleted_at: DateTime.t() | nil,
           inserted_at: DateTime.t() | nil,
           updated_at: DateTime.t() | nil
@@ -35,6 +37,7 @@ defmodule ExCubicIngestion.Schema.CubicTable do
     field(:name, :string)
     field(:s3_prefix, :string)
     field(:is_raw, :boolean)
+    field(:is_active, :boolean)
 
     field(:deleted_at, :utc_datetime)
 
@@ -46,13 +49,19 @@ defmodule ExCubicIngestion.Schema.CubicTable do
     from(table in __MODULE__, where: is_nil(table.deleted_at))
   end
 
+  @spec active :: Ecto.Queryable.t()
+  defp active do
+    from(table in __MODULE__, where: table.is_active and is_nil(table.deleted_at))
+  end
+
   @spec get_by!(Keyword.t() | map(), Keyword.t()) :: t() | nil
   def get_by!(clauses, opts \\ []) do
     Repo.get_by!(not_deleted(), clauses, opts)
   end
 
   @doc """
-  Given an enumerable of S3 prefixes, return those prefixes which represent a #{__MODULE__} and their table.
+  Given an enumerable of S3 prefixes, return those prefixes which represent a #{__MODULE__} and
+  their active table.
   """
   @spec filter_to_existing_prefixes(Enumerable.t()) :: [{String.t(), t()}]
   def filter_to_existing_prefixes(prefixes) do
@@ -67,7 +76,7 @@ defmodule ExCubicIngestion.Schema.CubicTable do
         |> Enum.to_list()
 
       query =
-        from(table in not_deleted(),
+        from(table in active(),
           where: table.s3_prefix in ^without_change_tracking
         )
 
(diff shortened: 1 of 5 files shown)


ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_table.ex
```diff
@@ -67,7 +76,7 @@ defmodule ExCubicIngestion.Schema.CubicTable do
         |> Enum.to_list()
 
       query =
-        from(table in not_deleted(),
+        from(table in active(),
```
grejdi-mbta, line 79: filter out any inactive prefixes/tables.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_table.ex
```diff
@@ -85,7 +94,7 @@ defmodule ExCubicIngestion.Schema.CubicTable do
   end
 
   @doc """
-  Get all active tables, including ODS snapshot for ODS tables.
+  Get all tables, including ODS snapshot for ODS tables.
```
grejdi-mbta, line 97: we'll pull everything here as it's being used during ingestion of loads, not at incoming step. allows loads that are still waiting to be ingested, but their table is inactive.

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_table.ex
```diff
@@ -26,6 +27,7 @@ defmodule ExCubicIngestion.Schema.CubicTable do
           name: String.t() | nil,
           s3_prefix: String.t() | nil,
           is_raw: boolean() | nil,
+          is_active: boolean() | nil,
```
paulswartz, line 30: question: What does `nil` mean here?
question: Should this be a "status" enum for extensibility rather than a boolean?
question: Should deleting a table set `is_active` (or the enum) to false?

ex_cubic_ingestion/test/ex_cubic_ingestion/oban_worker_error_test.exs
```diff
@@ -14,7 +14,8 @@ defmodule ExCubicIngestion.Schema.ObanWorkerErrorTest do
       table =
         Repo.insert!(%CubicTable{
           name: "cubic_dmap__sample",
-          s3_prefix: "cubic/dmap/sample/"
+          s3_prefix: "cubic/dmap/sample/",
+          is_active: true
```
paulswartz, line 18: question: rather than updating all the uses of this struct, can you set the default for `is_active` to true?

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_table.ex
grejdi-mbta: > question: What does nil mean here?

I based this off of the `is_raw` flag. I guess I can default it to `true` and that should answer your question.

> question: Should this be a "status" enum for extensibility rather than a boolean?

I don't know what other statuses we would incorporate, but in an effort to not over-engineer I opted for something more specific.

> question: Should deleting a table set is_active (or the enum) to false?

I don't think so. Also, I don't have a "delete" functionality right now. This field is there as a part of the default fields of database tables in general.

ex_cubic_ingestion/test/ex_cubic_ingestion/oban_worker_error_test.exs
grejdi-mbta: will do. opted for similar changes to `is_raw`

ex_cubic_ingestion/lib/ex_cubic_ingestion/schema/cubic_table.ex
grejdi-mbta: Added the defaulting to `true`

paulswartz: You also don't have a "set active to false" functionality. What are you thinking is the difference between a table that's deleted and table that's not active?

grejdi-mbta: > You also don't have a "set active to false" functionality.

This functionality will not be introduced until I have a "Admin" area for the Data Platform.

> What are you thinking is the difference between a table that's deleted and table that's not active?

A table that's deleted will no longer be resurrected. A table that's not active can be become active again.

paulswartz: Can't you set `deleted_at` back to `NULL` to un-delete a table?

grejdi-mbta: > Can't you set deleted_at back to NULL to un-delete a table?

I can, but I think it's better to keep active and deleted separate. I mentioned below that we only use active flag when filtering incoming, but not when ingesting. The idea is that, we might make a table inactive, but there are loads for that table that are in the process of being ingested and verified as valid, and therefore should finishing processing.

grejdi-mbta: So unfortunately because I have `INSERT`s in the migrations for tables, defaulting to `true` causes an issue with where the migrations can't complete due to the missing "is_active" field. I have reverted the default to `nil`. I also now just update and add the `is_active: true` to only the relevant tests, hence less changes.

paulswartz (approved):

PR #116, opened 2024-02-21
[elixir] feat: add admin app

grejdi-mbta:


grejdi-mbta:
@paulswartz Would love initial thoughts here.

admin/config/dev.exs
```diff
@@ -0,0 +1,10 @@
+import Config
+
+config :ex_aws,
+  # note: 'region' doesn't work. it's inherited from the profile
+  # region: [{:system, "AWS_REGION"}, {:awscli, "default", 30}, :instance_role],
+  access_key_id: [{:system, "AWS_ACCESS_KEY_ID"}, {:awscli, "default", 30}],
```
grejdi-mbta, line 6: I don't need these now, but will eventually.

admin/lib/admin/application.ex
```diff
@@ -0,0 +1,28 @@
...
+
+  use Application
+
+  @impl true
+  def start(_type, _args) do
+    children = [
+      {Admin.Repo, []},
+      {Plug.Cowboy,
+       scheme: :https,
+       options: [
```
grejdi-mbta, line 14: Question: I need to move this to a config so it can be different on prod, but not sure what the best way to go about it.

admin/lib/admin/router.ex
```diff
@@ -0,0 +1,71 @@
...
+  import Ecto.Query
+
+  require Logger
+
+  alias Admin.Repo
+  alias Admin.Schema.CubicTable
+
+  @template_dir "lib/admin/templates"
+
+  plug(Plug.SSL, rewrite_on: [:x_forwarded_proto, :x_forwarded_host, :x_forwarded_port])
```
grejdi-mbta, line 13: How do we still accept 'http'?

admin/lib/admin/router.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  get "/test" do
+    tables = Repo.all(from(table in CubicTable, where: is_nil(table.deleted_at)))
+
+    Logger.info("[admin] #{length(tables)}")
+
+    render(conn, "index.html")
+  end
+
+  get "/admin" do
```
grejdi-mbta, line 44: Would like Guardian w/ Keycloak. Any good implemtations to look at? MyCharlie is using it in combination with Phoenix.

admin/lib/admin/templates/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
```
grejdi-mbta, line 1: Thinking of using Tailwind and HTMX. Thoughts?

admin/lib/admin/application.ex
paulswartz:
```elixir
# config/config.exs
config :admin,
  cowboy_options: [ scheme: :https, ... ]

# lib/admin/application.ex
{Plug.Cowboy, Application.get_env(:admin, :cowboy_options)}
```

admin/lib/admin/schemas/cubic_ods_table_snapshot.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Admin.Schema.CubicOdsTableSnapshot do
```
paulswartz, line 1: question: these are the same tables that are maintained by the main application, correct? If so, having two different applications talking to the same database is definitely an anti-pattern.

admin/lib/admin/templates/index.html
paulswartz: Tailwind seems fine. I'd go with LiveView if you're already in the Elixir world, but I haven't heard many bad things about HTMX either.

paulswartz (commented):
I'm a little confused, and unfortunately there isn't much detail in the Asana ticket. 

Is this an application that a developer will run on their local machine to control the Data Platform tables? If it's expected to be deployed to AWS, then I wouldn't create a parallel application: instead I'd add Phoenix to the existing application and put a load balancer in front of it to handle HTTPS.

PR #126, opened 2025-02-28
[elixir] feat: remove invalid schema check and glue job

grejdi-mbta:
* We are turning off glue jobs at this time, as the output it generates are not in use. Most users of the data are looking in the DMAP Import database.
* We also turn off schema validation, as it is done in other pipelines prior to ingestion into the DMAP Import database, and soon ODIN.

### DIFF
--- a/ex_cubic_ingestion/lib/ex_cubic_ingestion/validate_incoming.ex
+++ b/ex_cubic_ingestion/lib/ex_cubic_ingestion/validate_incoming.ex
@@ -7,8 +7,6 @@ defmodule ExCubicIngestion.ValidateIncoming do
   use GenServer
 
   alias ExCubicIngestion.Schema.CubicLoad
-  alias ExCubicIngestion.Schema.CubicTable
-  alias ExCubicIngestion.SchemaFetch
 
   require Logger
   require Oban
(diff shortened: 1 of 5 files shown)


ex_cubic_ingestion/lib/ex_cubic_ingestion/validate_incoming.ex
```diff
@@ -56,41 +54,14 @@ defmodule ExCubicIngestion.ValidateIncoming do
...
+  def run(_state) do
     ready_loads = CubicLoad.get_status_ready()
 
-    # check schemas, and split into valid and invalid loads
-    {valid_ready_loads, invalid_ready_loads} =
-      Enum.split_with(ready_loads, &valid_schema?(state, &1))
-
     # start ingestion for those with valid schemas
-    Enum.each(valid_ready_loads, fn {load_rec, _table_rec} ->
+    Enum.each(ready_loads, fn {load_rec, _table_rec} ->
```
grejdi-mbta, line 61: Schema checks happens further down. In the DMAP-Import pipeline and soon the ODIN pipeline.

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex
```diff
@@ -26,53 +26,10 @@ defmodule ExCubicIngestion.Workers.Ingest do
...
-
-    lib_ex_aws.request(
-      ExAws.Glue.start_job_run(glue_job_name, %{
-        "--extra-py-files":
-          "s3://#{bucket_operations}/#{prefix_operations}packages/py_cubic_ingestion.zip",
-        "--ENV": Jason.encode!(env_payload),
-        "--INPUT": Jason.encode!(input_payload)
-      })
-    )
+    load_rec_ids
```
grejdi-mbta, line 30: Not much left to do here, but update the status to archive.

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex
```diff
@@ -106,7 +63,7 @@ defmodule ExCubicIngestion.Workers.Ingest do
       Application.fetch_env!(:ex_cubic_ingestion, :glue_job_cubic_ingestion_ingest_incoming)
 
     # pause a litte before getting status
-    Process.sleep(30_000)
+    Process.sleep(1)
```
grejdi-mbta, line 66: Making tests run faster. This functionality is no longer used, as in we don't call glue jobs any more.

paulswartz: question: if it's no longer used, should we remove the code and the tests?

paulswartz (commented):
question: if we don't have the validation or the Glue jobs, do we need this application at all?

grejdi-mbta:
> question: if we don't have the validation or the Glue jobs, do we need this application at all?

Yes. This application moves files from Incoming to Archive as soon as they land in Incoming. We plan to replace this part as well, but it's something that we felt OK about cutting from the scope of ODIN for the time being.

paulswartz:
if the only role of this application is to move the files from Incoming to Archive, can we write the files directly to Archive?

ex_cubic_ingestion/lib/ex_cubic_ingestion/workers/ingest.ex
grejdi-mbta: I went down this path a little bit, and it started ballooning, so decided to put in the lowest effort to get at the desired state.

grejdi-mbta:
> if the only role of this application is to move the files from Incoming to Archive, can we write the files directly to Archive?

Cubic/Qlik Replicate requires the ability to delete objects, so if we point Qlik to Archive, they might delete files on us. We need this 2-step process to make sure Cubic can't delete files on us. If you're saying that we should update this application to write to Archive, it is doing that. It's just that now it feels convoluted, due to removing the features that required the convoluted approach.

paulswartz:
Gotcha. Then this seems fine, once the tests pass.

PR #132, opened 2025-10-02
Adding token history table to data_platform

ealexa05:


ealexa05:
Redundant with https://github.com/mbta/data_platform/pull/133

PR #147, opened 2026-07-16
Add EC_STAGE.METRIC_HISTORY

MaxAlex:
Adds a new migration to start copying over EC_STAGE.METRIC_HISTORY.


ealexa05 (approved):

ealexa05:
Per Max, noting that CI seems to be broken here; will make sure to check that deploy works

